PR #5, opened 2020-06-14
V0.2.0

josepot:
I this PR:

- Removing `useObservable` from the API

After long process of consideration: I've reached the conclusion that it would be better to remove `useObservable` from the API. I think that if this library is used as it's intended, then there is no need to export it.

I'm more than willing to revisit this decision in the future if I see evidence (through non-contrived examples) that this hook is actually needed.

- New overload for `connectFactoryObservable`

Another change in v0.2.0 is the fact that the `getObservable` function of `connectFactoryObservable` can now take an `Object` as the instance key. I'm still not convinced about this new overload, but I can think of a few use-cases when it could be good to have it. Although, admittedly I've never encountered myself a situation when this is needed... So, am wondering if I should wait until we find real use-cases before adding it to the API? I don't know, on the other hand: it doesn't really hurt to have it? 

- Adding code of conduct

- We got a logo thanks to @clayforthcarr !! :tada: :bow: 
(Actually, I just dropped this commit from the PR and I added directly to `main`... I couldn't wait!)

### DIFF
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -1,0 +1,68 @@
+# Contributor Covenant Code of Conduct
+
+## Our Pledge
+
+In the interest of fostering an open and welcoming environment, we as contributors and maintainers
+pledge to making participation in our project and our community a harassment-free experience for
+everyone, regardless of age, body size, disability, ethnicity, sex characteristics, gender identity
+and expression, level of experience, education, socio-economic status, nationality, personal
+appearance, race, religion, or sexual identity and orientation.
+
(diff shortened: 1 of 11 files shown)


josepot:
@all-contributors please add @josepot for code, ideas, maintenance and test

josepot:
@all-contributors please add @voliva ideas and review

josepot:
@all-contributors please add @clayforthcarr for design

josepot:
closing this and reopening it because I'm getting rid of the `master` branch in favor of `main`

PR #9, opened 2020-06-14
V0.2.0

josepot:
I this PR:

- Removing `useObservable` from the API

After long process of consideration: I've reached the conclusion that it would be better to remove `useObservable` from the API. I think that if this library is used as it's intended, then there is no need to export it.

I'm more than willing to revisit this decision in the future if I see evidence (through non-contrived examples) that this hook is actually needed.

- New overload for `connectFactoryObservable`

Another change in v0.2.0 is the fact that the `getObservable` function of `connectFactoryObservable` can now take an `Object` as the instance key. I'm still not convinced about this new overload, but I can think of a few use-cases when it could be good to have it. Although, admittedly I've never encountered myself a situation when this is needed... So, am wondering if I should wait until we find real use-cases before adding it to the API? I don't know, on the other hand: it doesn't really hurt to have it? 

- Adding code of conduct

- We got a logo thanks to @clayforthcarr !! :tada: :bow: 
(Actually, I just dropped this commit from the PR and I added directly to `main`... I couldn't wait!)


test/connectFactoryObservable.test.tsx
```diff
@@ -41,4 +42,72 @@ describe("connectObservable", () => {
...
+      foo: "foo3",
+    }
+
+    let count = 0
+
+    const [useFoo] = connectFactoryObservable((obj: { foo: string }) =>
+      defer(() => {
+        count++
+        return of(obj.foo).pipe(map(text => [text, count].join(".")))
+      }),
```
voliva, line 63: That's a nice way of asserting recalculations. I like it 👍

josepot: :see_no_evil: the tests are horrible because they are not descriptive and they test way too much... However, they serve the purpose of making sure that I don't break anything major after every refactor, and as ugly as they are they have prevented me from some pretty shitty things already. Anyways, I'm currently working on a better set of tests.

josepot:
> I'm not really sure about this one... It would be nice if we can think of an example where this would be useful...

:100: I'm dropping the commit

PR #11, opened 2020-06-15
prevent surfacing errors happening in components to be unmounted

voliva:
Recently I've been thinking about a possible issue with the current re-rxjs bindings.

It has to do with possible false assumptions when using `connectFactoryObservable`. As an example, imagine we write a factory observable such as:

```ts
const [useUserPosts] = connectFactoryObservable(
  (userId: string) => user$.pipe(
    map(users => users[id]),
    map(user => user.posts),
    combineLatest(post$),
    map(([ids, posts]) => ids.map(id => posts[id]))
)
```

We might make the assumption that, as we will call `useUserPosts()` in components that receive an existing `userId`, we don't need to check whether that userId exists (again, because we're going to call this hook with existing ids).

The problem comes when a user is removed. In this logic, no error should be thrown because as the component gets unmounted, it won't be requesting an id that doesn't exist, but that's not the case because the component will get unmounted _after_ this value has computed, so the code above will throw an uncaught `can't access property 'posts' of undefined`.

Here I have a [sandbox with a simplified todomvc](https://0pkl9.codesandbox.io/) that shows the current behaviour.

_👇 Outdated, see comments below_
> This PR fixes this by catching all the errors that happen in the stream, and throwing them in a `useEffect`. This way, if react has unmounted the component, the error won't be thrown.
> 
> I don't fully like the test I've written, but it works so far. It was mostly trial-and-error until I found something that failed without the fix and works fine after it. I'll try to find something a bit better later on if needed.
> I'd also want to add a test to check that other errors are still being thrown, but I can't because any uncaught errors will cause jest to fail the test (see facebook/jest#5620).

### DIFF
--- a/src/useObservable.ts
+++ b/src/useObservable.ts
@@ -1,7 +1,15 @@
 import { useEffect, useReducer } from "react"
 import { BehaviorObservable, SUSPENSE } from "./"
 
-const reducer = (_: any, action: any) => action
+const reducer = (
+  _: any,
+  { type, payload }: { type: "next" | "error"; payload: any },
+) => {
+  if (type === "error") {
+    throw payload
+  }
+  return payload
+}
 const init = (source$: BehaviorObservable<any>) => {
   try {
     return source$.getValue()
@@ -18,11 +26,28 @@ export const useObservable = <O>(
 
   useEffect(() => {
     try {
-      dispatch(source$.getValue())
+      dispatch({
+        type: "next",
+        payload: source$.getValue(),
+      })
     } catch (e) {
-      dispatch(SUSPENSE)
+      dispatch({
+        type: "next",
+        payload: SUSPENSE,
+      })
     }
-    const subscription = source$.subscribe(dispatch)
+    const subscription = source$.subscribe(
+      value =>
+        dispatch({
+          type: "next",
+          payload: value,
+        }),
+      error =>
+        dispatch({
+          type: "error",
+          payload: error,
+        }),
+    )
     return () => subscription.unsubscribe()
   }, [source$, unsubscribeGraceTime])
 
--- a/test/connectFactoryObservable.test.tsx
+++ b/test/connectFactoryObservable.test.tsx
@@ -1,6 +1,6 @@
 import { connectFactoryObservable } from "../src"
-import { from, of, defer, concat } from "rxjs"
-import { renderHook } from "@testing-library/react-hooks"
+import { from, of, defer, concat, BehaviorSubject } from "rxjs"
+import { renderHook, act } from "@testing-library/react-hooks"
 
 const wait = (ms: number) => new Promise(res => setTimeout(res, ms))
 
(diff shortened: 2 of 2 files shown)


josepot:
Hi @voliva 

Thanks a lot for finding this issue and for coming up with a strategic solution. However, the following comment: 
> I don't fully like the test I've written, but it works so far. It was mostly trial-and-error until I found something that failed without the fix and works fine after it. I'll try to find something a bit better later on if needed.
> I'd also want to add a test to check that other errors are still being thrown, but I can't because any uncaught errors will cause jest to fail the test (see facebook/jest#5620).

Made me think that perhaps the issue was that since the error is thrown from a `useEffect` callback, then it's possible that React can't catch it. If that was the case, then errors from the streams couldn't be caught through an error-boundary component... Meaning that it's likely that we currently have that issue and I think that addressing that issue will in its turn solve the one that you found.

So, I did some research and TIL that throwing an error from a hook has to be done [more or less like this.](https://github.com/facebook/react/issues/14981#issuecomment-468460187)

So, there is no need for that `useEffect` after all :sweat_smile: Therefore, I would prefer it if the reducer looked like this:
```ts
const reducer = (                                              
  _: any,                                                      
  { type, payload }: { type: "error" | "value"; payload: any },
) => {                                                         
  if (type === "error") {                                      
    throw payload                                              
  }                                                            
  return payload                                               
}                                                              
```

And then we would be able to have the tests that you were looking for:

```ts
it("errors can be caught in error boundaries", () => {                                    
  const errStream = new BehaviorSubject(1)                                                
  const [useError] = connectFactoryObservable(() => errStream)                            
                                                                                          
  renderHook(() => useError())                                                            
                                                                                          
  expect(() =>                                                                            
    act(() => {                                                                           
      errStream.error("error")                                                            
    }),                                                                                   
  ).toThrow()                                                                             
})                                                                                        
                                                                                          
it("doesn't throw errors on components that will get unmounted on the next cycle", () => {
  const errStream = new BehaviorSubject(1)                                                
  const [useError] = connectFactoryObservable(() => errStream)                            
                                                                                          
  const { unmount } = renderHook(() => useError())                                        
                                                                                          
  expect(() =>                                                                            
    act(() => {                                                                           
      errStream.error("error")                                                            
      unmount()                                                                           
    }),                                                                                   
  ).not.toThrow()                                                                         
})                                                                                                                                                                             
```

josepot:
@all-contributors please add @voliva for code and test

josepot:
@all-contributors please add @josepot for review

voliva:
That's so interesting. I kind of assumed the reducer in `useReducer` would execute as soon as you make the dispatch, but I see that's not the case.

Seems like react calls the reducer when it's updating the component, so if the component is unmounted it won't throw the error. Nice catch!

josepot:
@all-contributors please add @voliva for code and test

PR #15, opened 2020-06-16
Docs: add new example to the README

josepot:
I'm not sure about this, but I'm thinking about adding a kinda complete example on the README until the first version of the docs is ready... What do you think @voliva ?

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,10 +29,6 @@ Main features
 - :microscope: Tiny and tree-shakeable
 - :muscle: Supports TypeScript
 
-## Examples
-
-- [This is a contrived example](https://codesandbox.io/s/crazy-wood-vn7gg?file=/src/fakeApi.js) based on [this example](https://reactjs.org/docs/concurrent-mode-patterns.html#reviewing-the-changes) from the React docs.
-
 ## Docs
 ```tsx
 const [useDocs] = connectObservable(NEVER)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -54,7 +50,113 @@ function App() {
...
+    </div>
+  )
+}
+
+export default function App() {
+  return (
+    <>
+      <Header>Search Github Repos</Header>
+      <Search onSubmit={onSubmit} />
+      <Suspense fallback={() => <LoadingResults />}>
```
voliva, line 152: The fallback should be an element. I also see there's the same mistake in the temporary Docs section.

README.md
```diff
@@ -29,10 +29,6 @@ Main features
 - :microscope: Tiny and tree-shakeable
 - :muscle: Supports TypeScript
 
-## Examples
-
-- [This is a contrived example](https://codesandbox.io/s/crazy-wood-vn7gg?file=/src/fakeApi.js) based on [this example](https://reactjs.org/docs/concurrent-mode-patterns.html#reviewing-the-changes) from the React docs.
-
 ## Docs
 ```tsx
 const [useDocs] = connectObservable(NEVER)
```
voliva, line 34: Now that I think of, ☝️ this kinda reads as docs will NEVER come :P

README.md
josepot: :scream:

README.md
josepot: Yep... I'm not sure if ppl will find that funny :grimacing: So, we better hurry up and finish the first version of the docs.

PR #17, opened 2020-06-16
Remove React-(dom/native) unstable_batchedUpdates

josepot:
I've been thinking a lot about this lately and I'm almost convinced that it doesn't make sense for us to use the `unstable_batchedUpdates`... The reason being that those changes that come from within react event-handlers (e.g. `onClick`) will be batched already. On the other hand, those updates that come from external sources (i.e: APIs whether they are socket-based or REST) the recommendation would be to observer them through the `asapScheduler`. In other words, it would be up to the user to decide the priority of each "event", but if this library is used as it's intended, then observing changes through the `asapScheduler` should remove the need for prematurely batching external updates.

Anyways, this is a decision that can be revisited in the future and it would only imply a patch upgrade. So, I much rather not to fall for what now I think it could be a premature optimization.

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,14 +22,6 @@
     "react": ">=16.8.0",
     "rxjs": ">=6"
   },
-  "peerDependenciesMeta": {
-    "react-dom": {
-      "optional": true
-    },
-    "react-native": {
-      "optional": true
-    }
-  },
   "husky": {
     "hooks": {
       "pre-commit": "tsdx lint"
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -1,7 +1,3 @@
-import { setBatch } from "./utils/batch"
-import { unstable_batchedUpdates as batch } from "./utils/react-batched-updates"
-setBatch(batch as any)
-
 export const SUSPENSE = Symbol("SUSPENSE")
 export { BehaviorObservable } from "./BehaviorObservable"
 export { connectObservable } from "./connectObservable"
--- a/src/operators/distinct-share-replay.ts
+++ b/src/operators/distinct-share-replay.ts
@@ -1,6 +1,5 @@
 import { Observable, Subscription, Subject } from "rxjs"
 import { SUSPENSE, BehaviorObservable } from "../"
-import { getBatch } from "../utils/batch"
 
 function defaultTeardown() {}
 
@@ -13,7 +12,6 @@ export const distinctShareReplay = <T>(
   let subscription: Subscription | undefined
   let refCount = 0
   let currentValue: { value: T }
-  const batch = getBatch()
 
   const result = new Observable<T>(subscriber => {
     refCount++
@@ -26,9 +24,7 @@ export const distinctShareReplay = <T>(
             currentValue.value === EMPTY_VALUE ||
             !compareFn(value, currentValue.value)
           ) {
-            batch(() => {
-              subject!.next((currentValue.value = value))
-            })
+            subject!.next((currentValue.value = value))
           }
         },
         error(err) {
--- a/src/utils/batch.ts
+++ b/src/utils/batch.ts
@@ -1,8 +1,0 @@
-function defaultNoopBatch(callback: () => void) {
-  callback()
-}
-
-let batch = defaultNoopBatch
-
-export const setBatch = (newBatch: () => void) => (batch = newBatch)
-export const getBatch = () => batch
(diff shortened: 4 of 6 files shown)


voliva:
> On the other hand, those updates that come from external sources (i.e: APIs whether they are socket-based or REST) the recommendation would be to observer them through the `asapScheduler`

I don't fully understand this. From this [sandbox](https://tf5oo.csb.app/) that looks at how batching is done when handled through an event vs through a promise callback, I see that from the promise callback the state changes are not batched, even though they are called in the same event cycle. So doesn't that mean that `asapScheduler` will have no effect on that? Worded in another way, by observing on `asapScheduler` it seems like setStates won't be batched, right?
I don't fully understand how batching works in React.

However
> Anyways, this is a decision that can be revisited in the future and it would only imply a patch upgrade. So, I much rather not to fall for what now I think it could be a premature optimization.
That's a totally valid point. If it proves that's negatively effecting the performance in large-scale apps, it can always be added as an enhancement.

josepot:
Let's say that we are using React `16.13.1` and that we have [the following code on a `React.StrictMode` app](https://codesandbox.io/s/adoring-night-2ru26?file=/src/App.js):

```tsx
import React, { useState, useEffect } from "react";

const cbs = [];
const registerSetCount = cb => {
  cbs.push(cb);
  if (cbs.length === 2) {
    console.log("starting count-down");
    setTimeout(() => {
      for (let i = 0; i < 2; i++) {
        cbs[i](x => x + 1);
      }
    }, 1000);
  }
};

function Parent() {
  let [count, setCount] = useState(0);
  useEffect(() => {
    console.count("Parent render");
  });
  useEffect(() => {
    registerSetCount(setCount);
  }, []);
  return (
    <div>
      Parent: {count}
      <Child />
    </div>
  );
}

function Child() {
  let [count, setCount] = useState(0);
  useEffect(() => {
    console.count("Child render");
  });
  useEffect(() => {
    registerSetCount(setCount);
  }, []);
  return <div>Child: {count}</div>;
}

export default function App() {
  return <Parent />;
}
```

If we don't use `unstable_batchedUpdates` then the child will render twice as a result of the update that comes from the `setTimeout`.

However, on the same version of React and also on `StrictMode` [if we wrap the `for` loop on a `unstable_batchedUpdates` callback](https://codesandbox.io/s/flamboyant-framework-eorez?file=/src/App.js), then the child will only render once... Like if that update had been triggered on an `onClick` react event-handler. That's because on the render phase react batches all the updates before rendering the components.

And lastly, if [we use React experimental, then on the original code (without the `unstable_batchedUpdates`)](https://codesandbox.io/s/quiet-wood-5fek6?file=/src/App.js) the child also gets rendered only once :exploding_head:. That's because on concurrent-mode, all the state-updates that get triggered "together" get batched by default.

So, what I meant to say is that on Concurrent mode (React experimental) using the `asapScheduler` would "do the trick". Because all the state-updates would get sequentially executed at the beginning of the next event-loop and therefore React will automatically batch them. I think that's one of the reasons why we don't see tearing on https://github.com/dai-shi/will-this-react-global-state-work-in-concurrent-mode/pull/18

So, I guess that we could provide a version of the `asapScheduler` that would take a "batch" function as an argument... That way if you are using on react-dom or on react-native, you pass it to us and the operator would run the `observer.next` calls wrapped into that function. I rather doing that than baking `unstable_batchedUpdates` on the library

voliva:
Wow.... would it be possible to expose it as a simple operator? Imagine something like

```ts
const [useLastPrice] = connectObservable(getLastPrices().pipe(
    batchReactUpdates()
))
```

Thinking it better this operator kind of feels like it should either always be there, or never. In that case it's probably better to just reconsider the whole batching thing later on.

josepot:
It would be possible, yes... However `unstable_batchedUpdates` is... unstable :smile: and one day won't be there. Also, most importantly it doesn't come from `React`. It comes from 'react-dom', 'react-native' and I don't even want to know if other react renderers have their own version...

That's why I would rather to have a package like `@re-rxjs/react-dom` that would expose the `batchUpdates` operator for `react-dom` and `@re-rxjs/react-native` would expose the one for react-native, etc... That way the core library can keep the same API for both concurrent, strict and legacy mode. But this is part of a larger discussion...

Anyways, I think that we agree that for now it makes sense to merge this. Thanks a lot for your feedback @voliva !

PR #23, opened 2020-06-22
Improve createInput types and behavior

josepot:
I've been thinking about this as I was writing the docs and I've realized that `createInput` should also allow for "void" inputs... So, I think that this is the behavior that I would like to have on `createInput`. @voliva thoughts?


josepot:
> but it's something that the consumer can add when needed

In fact, it's something that will get automatically added later on when they connect the observable :slightly_smiling_face:

PR #25, opened 2020-06-22
Initial docs

josepot:
There is still a ton of room for improvement, but it's slightly better than what we currently have. So, lets take it one step at the time...

That being said, @voliva could you please help me improve this?


README.md
```diff
@@ -33,23 +33,158 @@ Main features
...
+
+  return (
+    <article>
+      <h1>{story.title}</h1>
+      <p>{story.description</p>
+    </article>
+  )
+}
+```
+Returns a hook that receives as arguments the same arguments as the factory function
```
bhavesh-desai-scratch, line 86: Returns a hook that presents the latest value of an observable generated from a factory function.

Accepts: 
A factory function that returns an Observable

Returns [1 , 2]:
1. A function with the same parameters as the factory function, calling this function in a React component will create and subscribe to an observable generated by the accepted factory function. The function will then return the latest update from the observable.
2. A shared replayable version of the observable generated by the factory function that is used by the hook internally.

bhavesh-desai-scratch: need to name 1 and 2!

README.md
```diff
@@ -33,23 +33,158 @@ Main features
...
+### connectObservable
+```ts
+const [useCounter, sharedCounter$] = connectObservable(
+  clicks$.pipe(
+    scan(prev => prev + 1, 0),
+    startWith(0),
+  )
+)
+```
+Returns a hook that provides the latest update of the observable, and an enhanced
```
bhavesh-desai-scratch, line 65: Returns a hook that provides the latest update of the **accepted** observable, and an enhanced

README.md
```diff
@@ -33,23 +33,158 @@ Main features
...
+Returns a hook that receives as arguments the same arguments as the factory function 
+and provides the latest update of the observable returned by the factory function.
+
+It also returns a function that provides the shared observable that the hook uses, 
+which is shared and "replayable". That observable is disposed when its `refCount`
+goes down to zero.
+
+### distinctShareReplay
+
+A RxJS pipeable operator which performs a custom `shareReplay` that can be very useful
```
bhavesh-desai-scratch, line 95: Lose "very"

Should we describe why it is useful? 
1: It ensures React components only update when necessary 
2: Components using the same hooks use the same underlying Observables
3: Unsubscribes to the underlying observable when all components using a hook are unmounted.

Though this might be too much. Just an Idea.

josepot:
> It might be worth putting the API docs in a specific format, such as:
> 
> Description:
> 
> When this is useful:
> 
> Returns:
> 
> Parameters:

Most definitely! However, the only purpose of this PR is to have something more "decent" than what we currently have. So, I've taken most of your feedback and I'm merging this for now because I can't sleep at night knowing that the "coming soon" joke is still no the "Docs" section of the current README.

josepot:
@all-contributors please add @bhavesh-desai-scratch for review

Issue #20, opened 2020-06-22
Test improvements


voliva:
Most of the work is already done, but there's still margin for improvement on the tests. I focused on those methods exposed as part of the API.

## connectObservable

* [x] It's missing a test for updates on the source stream.
* [x] Q: I understand re-rxjs doesn't do any sort of batching. Is there any way to show how to make react batch updates?
* [x] Maybe the test `it works with suspense` can be broken down. I haven't taken a deep look and I'm a bit out of the loop from suspense, but it will be interesting to see if we can break that test down.

## connectFactoryObservable

* [x] On test `it returns the latest emitted value` it only checks for the initial value. Needs to add in some updates on the source stream.
* [x] I think it would be good to have a test `it shares the subscription among all of the subscribers of the same key`. It is superseeded by another test, but this will document a bit better how the key works.
```ts
    /**
     * - one hook subscribes to key1
     * - it receives the initial value
     * - perform a change on key1 subject
     * - another hook subscribes to key1
     * - both receive the change made on key1 subject
     * - another hook subscribes to key2, and receives the initial value
     * - all hooks unmount
     */
```
* [x] The test `observable > if the source observable completes it keeps emitting the latest value until there are no more subscriptions` has an `expect` I found confusing:
```ts
  let diff = -1
  const [, getShared] = connectFactoryObservable((_: number) => {
    diff++
    return from([1, 2, 3, 4].map(val => val + diff))
  })

  let latestValue1: number = 0
  let nUpdates = 0
  const sub1 = getShared(0).subscribe(x => {
    latestValue1 = x
    nUpdates += 1
  })
  expect(latestValue1).toBe(4)
  expect(nUpdates).toBe(1) // <--- this one
  /**
   * Feels like it shouldn't be expected to receive only the last value from the source
   * as it seems like it's sort of debouncing the updates.
   * I've checked and it's not, it's only on this specific case where you have a synchronous
   * source that you will get the last value of that synchronous batch on the first subscription.
   * Following "synchronous batches" are fine (they will emit every single item), it's just
   * the initial one.
   * Maybe it's worth having a test just to have this case documented, even if the behaviour
   * changes later on if we consider this needs to be fixed.
   */
```
* [x] The error handling ones we mentioned how having a test working with react would be cool. I also wonder how to replicate this test also for `connectObservable` (as it applies as well)

## createInput

I found the test cover everything I could think of 👍 

## rxjs operators

[...]

PR #26, opened 2020-06-23
Fix: distinctShareReplay shouldn't skip values on sync subscriptions

josepot:
This addresses one of the issues found by @voliva on #20 (Thanks a lot for pointing that out BTW!)

Is this living proof that marble-tests can't be pretty useless for some stuff? Why can't I find a way to add a marble test that ensures this behavior?


josepot:
Guess what? I [inherited this bug from RxJS!](https://codesandbox.io/s/gifted-meitner-7mzhc?file=/src/index.ts)

PR #29, opened 2020-06-23
Test compare option in connectObservable

bhavesh-desai-scratch:


test/connectObservable.test.tsx
```diff
@@ -35,6 +35,35 @@ describe("connectObservable", () => {
...
+    expect(result.current.value).toEqual(0)
+
+    stream$.next({ value: 0, valueToIgnore: "B" })
+    rerender()
+    //should not update to the latest value in the stream
+    expect(result.current.valueToIgnore).toEqual("A")
+    expect(result.current.value).toEqual(0)
+
+    stream$.next({ value: 1, valueToIgnore: "B" })
+    rerender()
```
voliva, line 62: Small detail - I think `act` would be more semantic instead of `rerender` -

e.g.
```ts
act(() => {
    stream$.next({ value: 1, valueToIgnore: "B" })
})
```

voliva: Now that I think of, in fact, it would be awesome if we can test that the component actually doesn't rerender if `valueToIgnore` changes and `value` doesn't. It's something that would add more value.

test/connectObservable.test.tsx
```diff
@@ -35,6 +35,35 @@ describe("connectObservable", () => {
...
+
+    const compare = (a: TestUpdate, b: TestUpdate) => a.value === b.value
+    const [useLatestValue] = connectObservable(stream$, { compare })
+
+    const { result, rerender } = renderHook(() => useLatestValue())
+    expect(result.current.valueToIgnore).toEqual("A")
+    expect(result.current.value).toEqual(0)
+
+    stream$.next({ value: 0, valueToIgnore: "B" })
+    rerender()
```
josepot, line 56:
```suggestion
    act(() => {
      stream$.next({ value: 0, valueToIgnore: "B" })
    })
```

The "proper" way to do this stuff with `@testing-library/react-hooks` is to wrap the `stream@.next` on an `act` callback and not to call `rerender()`

test/connectObservable.test.tsx
josepot, line 62:
```suggestion
    act(() => {
      stream$.next({ value: 1, valueToIgnore: "B" })
    })
```
Same

test/connectObservable.test.tsx
```diff
@@ -1,14 +1,14 @@
+import { fireEvent, render, screen } from "@testing-library/react"
+import { renderHook } from "@testing-library/react-hooks"
```
josepot, line 2:
```suggestion
import { renderHook, act } from "@testing-library/react-hooks"
```
lets not forget to import `act` :slightly_smiling_face:

josepot:
@all-contributors please add @bhavesh-desai-scratch for tests

PR #30, opened 2020-06-23
docs: add usage examples to every method

voliva:
Forgive me I was late on reviewing the initial docs PR - Here are my suggestions that I think it improves them a bit.

I understand that those were initial docs, to have something better than a placeholder, so I agree that maybe this PR should be closed instead and this will be addressed when working on the final docs. But in case they might help, here they are.

Addresses two main points:
1. I didn't fully like adding keywords that people might not be familiar with. Specifically `shared` and `replayable`. So I rephrased that.
2. Some of the operators had the actual implementation in them. Although I can see how this is useful to get the point across (e.g. `distinctShareReplay`), I felt like in some places it was redundant (if people want to see the how it's implemented it's just 2 clicks away), and it would be of more helpful to show a usage example instead.

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,9 +62,12 @@ const [useCounter, sharedCounter$] = connectObservable(
   )
 )
 ```
-Returns a hook that provides the latest update of the accepted observable, and an enhanced 
-version of the original observable (the one that the hook will use), which is shared 
-and "replayable".
+Returns a hook that provides the latest update of the accepted observable, and the
+underlying enhanced observable, which shares the subscription to all of its subscribers,
+and always emits the latest value when subscribing to it.
+
+The shared subscription is closed as soon as there are no subscribers to that
+hook/observable.
 
 ### connectFactoryObservable
 ```tsx
@@ -96,6 +99,13 @@ can be used for composing other streams that depend on it. This observable is di
 when its `refCount` goes down to zero.
 
 ### distinctShareReplay
+```ts
+const activePlanetName$ = planet$.pipe(
+  filter(planet => planet.isActive),
+  map(planet => planet.name),
+  distinctShareReplay()
+)
+```
 
 A RxJS pipeable operator which performs a custom `shareReplay` that can be useful 
 when working with these bindings. It's roughly the equivalent of:
@@ -112,40 +122,54 @@ const distinctShareReplay = <T>(compare?: Object.is) => (
 ```
 
 The enhanced observables returned from `connectObservable` and `connectFactoryObservable` 
-have been enhanced like this:
-
-```ts
-const shared$ = concat(original$, NEVER).pipe(distinctShareReplay())
-```
+have been enhanced with this operator.
 
 ### SUSPENSE
 
-This is a special symbol that can be emitted from our observables to let the react-hook
+```ts
+const story$ = selectedStoryId$.pipe(
+  switchMap(id => concat(
+    SUSPENSE,
+    getStory$(id)
+  ))
+)
+```
+
+This is a special symbol that can be emitted from our observables to let the react hook
 know that there is a value on its way, and that we want to leverage React Suspense API
 while we are waiting for that value.
 
 ### suspend
 
-A RxJS creation operator that performs a `startWith(SUSPENSE)` on the source observable.
-
 ```ts
-const suspend = <T>(source$: ObservableInput<T>) =>
-  from(source$).pipe(startWith(SUSPENSE))
+const story$ = selectedStoryId$.pipe(
+  switchMap(id => suspend(getStory$(id))
+)
 ```
 
+A RxJS creation operator that prepends a `SUSPENSE` on the source observable.
+
 ### suspended
+
+```ts
+const story$ = selectedStoryId$.pipe(
+  switchMap(id => getStory$(id).pipe(
+    suspended()
+  ))
+)
+```
 
 The pipeable version of `suspend`
 
 ### switchMapSuspended
 
-Like `switchMap` but applying a `startWith(SUSPENSE)` to the inner observable.
-
 ```ts
-const switchMapSuspended = <Input, Output>(
-  fn: (input: Input) => ObservableInput<Output>,
-) => (src$: Observable<Input>) => src$.pipe(switchMap(x => suspend(fn(x))))
+const story$ = selectedStoryId$.pipe(
+  switchMapSuspended(id => getStory$(id))
+)
 ```
+
+Like `switchMap` but applying a `startWith(SUSPENSE)` to the inner observable.
 
 ### createInput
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -62,9 +62,12 @@ const [useCounter, sharedCounter$] = connectObservable(
   )
 )
 ```
-Returns a hook that provides the latest update of the accepted observable, and an enhanced 
-version of the original observable (the one that the hook will use), which is shared 
-and "replayable".
+Returns a hook that provides the latest update of the accepted observable, and the
```
bhavesh-desai-scratch, line 65: I actually prefered it before. We should assume users of this library know RX. Trying to describe this function in terms of subscribers and subscriptions is hard to read and a bit confusing IMO.

bhavesh-desai-scratch:
> 1. I didn't fully like adding keywords that people might not be familiar with. Specifically `shared` and `replayable`. So I rephrased that.


I actually disagree here. I think using RX terminology is far more precise and easier to understand. We could add links to these terms when we use them.

voliva:
> > 1. I didn't fully like adding keywords that people might not be familiar with. Specifically `shared` and `replayable`. So I rephrased that.
> 
> I actually disagree here. I think using RX terminology is far more precise and easier to understand. We could add links to these terms when we use them.

My point is that, even for people familiar to RX, I think it's unclear what `shared` and `replayable` mean. `shared` actually [has](https://rxjs-dev.firebaseapp.com/api/operators/share) [many](https://rxjs-dev.firebaseapp.com/api/operators/publish) [different](https://rxjs-dev.firebaseapp.com/api/operators/multicast) [ways](https://rxjs-dev.firebaseapp.com/api/operators/refCount) that behave slightly different to each other.

I think `replayable observable` is also a term we used to define this but it's not exactly something used widely in rxjs circles. There are many `replay` operators, but again, each one behaves slightly different.

I would see another alternative by using the operator name `shareReplay`, but explaining how our operator is different from that one, but I think it's more straight-forward to just describe what it does.

I think it's important at least to describe exactly how this `shared` and `replayable` actually work, as otherwise there are many assumptions to be made when working with this `connectObservable` (how are subscriptions cleaned up? does it unsubscribe and resubscribe? how many items does it replay?).

README.md
```diff
@@ -62,9 +62,12 @@ const [useCounter, sharedCounter$] = connectObservable(
...
+underlying enhanced observable, which shares the subscription to all of its subscribers,
+and always emits the latest value when subscribing to it.
+
+The shared subscription is closed as soon as there are no subscribers to that
+hook/observable.
```
josepot, line 70: Yep, this description is a lot more accurate than the previous one. I like it.

The only thing that I would change is the end of the last sentence: instead of `hook/observable` let's just say: "observable".

README.md
```diff
@@ -96,6 +99,13 @@ can be used for composing other streams that depend on it. This observable is di
 when its `refCount` goes down to zero.
 
 ### distinctShareReplay
+```ts
+const activePlanetName$ = planet$.pipe(
+  filter(planet => planet.isActive),
+  map(planet => planet.name),
+  distinctShareReplay()
+)
+```
```
josepot, line 108: This is awesome! I wanted to start each section with a clear and concise example, but then I got lazy. I like this.

README.md
```diff
@@ -112,41 +122,55 @@ const distinctShareReplay = <T>(compare?: Object.is) => (
...
+### SUSPENSE
 
 ```ts
-const shared$ = concat(original$, NEVER).pipe(distinctShareReplay())
+const story$ = selectedStoryId$.pipe(
+  switchMap(id => concat(
+    SUSPENSE,
+    getStory$(id)
+  ))
+)
```
josepot, line 135: Same, thanks!

README.md
```diff
@@ -112,41 +122,55 @@ const distinctShareReplay = <T>(compare?: Object.is) => (
...
 
-Like `switchMap` but applying a `startWith(SUSPENSE)` to the inner observable.
-
 ```ts
-const switchMapSuspended = <Input, Output>(
-  fn: (input: Input) => ObservableInput<Output>,
-) => (src$: Observable<Input>) => src$.pipe(switchMap(x => suspend(fn(x))))
+const story$ = selectedStoryId$.pipe(
+  switchMapSuspended(id => getStory$(id))
+)
```
josepot, line 169: I love how you have been building it up so that the user can appreciate the value of this "sugar" operator... That's precisely why I think that for this example it would be worth it to use a point-free style example:

```ts
const story$ = selectedStoryId$.pipe(
  switchMapSuspended(getStory$)
)
```

In other examples I have avoided using point-free style declarations (like in `connectFactoryObservable`) because I wanted to make sure that the user understood the parameters that the function was receiving. However, in this particular case and taking into account that the docs have been building up this momentum... I think that we must :smile:

README.md
```diff
@@ -112,41 +122,55 @@ const distinctShareReplay = <T>(compare?: Object.is) => (
...
 ### switchMapSuspended
 
-Like `switchMap` but applying a `startWith(SUSPENSE)` to the inner observable.
-
 ```ts
-const switchMapSuspended = <Input, Output>(
-  fn: (input: Input) => ObservableInput<Output>,
-) => (src$: Observable<Input>) => src$.pipe(switchMap(x => suspend(fn(x))))
+const story$ = selectedStoryId$.pipe(
+  switchMapSuspended(id => getStory$(id))
```
josepot, line 168:
```suggestion
  switchMapSuspended(getStory$)
```

josepot:
@all-contributors please add @voliva for docs

@all-contributors please add @bhavesh-desai-scratch for docs

@all-contributors please add @josepot for docs

josepot:
@all-contributors please add @voliva for docs

josepot:
@all-contributors please add @bhavesh-desai-scratch for docs

josepot:
@all-contributors please add @josepot for docs

josepot:
@all-contributors please add @josepot for docs

README.md
josepot: TBH I was having a very hard time when I wrote that to find a precise and concise way to describe how the returned observable is being enhanced... At some point I just got lazy and I wrote that knowing that it was a lazy way to describe it. It's lazy because there are sooo many different ways to "share" and/or "replay" with RxJS... What I had in my head was: "well it's ok, because later on in the section for `distinctShareReplay` I will extend this explanation".

Anyways, I like @voliva explanation a lot better because while being pretty concise, he actually explains really well how exactly it is that the enhanced observable is sharing/replaying.

Can that description be confusing the first time that you read it? Quite possible. Is it accurate? Yep, it is. So, I much rather for the reader to read it a few times until it sinks than to provide a handwavy explanation, which is what I think that we got before.

For me the ideal docs are the ones that start providing good examples so that anyone can grok things, but that also provide the necessary detail so that experts can understand what's actually going on.

All that being said, I still think that there is lots of room for improvement. So, let's keep bringing good ideas and suggestions on how to improve the docs.

PR #31, opened 2020-06-23
test: connectFactoryObservable updates, subscriptions and errors.

voliva:
These covers the 3 remaining points identified for connectFactoryObservable in #20

### DIFF
--- a/test/connectFactoryObservable.test.tsx
+++ b/test/connectFactoryObservable.test.tsx
@@ -1,6 +1,10 @@
 import { connectFactoryObservable } from "../src"
-import { from, of, defer, concat, BehaviorSubject } from "rxjs"
-import { renderHook, act } from "@testing-library/react-hooks"
+import { from, of, defer, concat, BehaviorSubject, throwError } from "rxjs"
+import { renderHook, act as actHook } from "@testing-library/react-hooks"
+import { render, act } from "@testing-library/react"
+import { map, switchMap } from "rxjs/operators"
+import { Component, ErrorInfo, FC } from "react"
+import React from "react"
 
 const wait = (ms: number) => new Promise(res => setTimeout(res, ms))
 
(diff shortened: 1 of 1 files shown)


test/connectFactoryObservable.test.tsx
```diff
@@ -1,14 +1,21 @@
...
 
 const wait = (ms: number) => new Promise(res => setTimeout(res, ms))
 
 describe("connectFactoryObservable", () => {
   const originalError = console.error
   beforeAll(() => {
     console.error = (...args: any) => {
-      if (/Consider adding an error/.test(args[0])) {
+      if (
+        /Uncaught 'controlled error'/.test(args[0]) ||
```
voliva, line 16: That's interesting - React stills throws the error (so it shows up as an uncaught exception in the console) even if there's an ErrorBoundary. Not only that, but CRA will overlay a ugly error message on the whole body if it's in dev mode, telling you that you should fix that error that happened.

It only uses the ErrorBoundary to handle that specific component crashing, and in prod it doesn't show the overlayed message.

test/connectFactoryObservable.test.tsx
```diff
@@ -20,9 +27,43 @@ describe("connectFactoryObservable", () => {
   })
   describe("hook", () => {
     it("returns the latest emitted value", async () => {
-      const [useNumber] = connectFactoryObservable((id: number) => of(id))
+      const valueStream = new BehaviorSubject(1)
```
bhavesh-desai-scratch, line 30: You could simplify this by removing the ID from the output. It is not necessary for this test.

test/connectFactoryObservable.test.tsx
```diff
@@ -20,9 +27,43 @@ describe("connectFactoryObservable", () => {
...
-      expect(result.current).toBe(1)
+      expect(result.current).toBe("1:1")
+
+      actHook(() => {
+        valueStream.next(3)
+      })
+      expect(result.current).toBe("1:3")
+    })
+
+    it("shares the subscription among all of the subscribers of the same key", async () => {
```
bhavesh-desai-scratch, line 43: Is the key not an implementation detail of `createFactoryObservable`?

Should this not be something like:

Multiple components using a hook returned from `createFactoryObservable`  with the same parameters share a multicasted subscription?

voliva: Yep, that's a better test description 👍

test/connectFactoryObservable.test.tsx
voliva: Good catch!

PR #39, opened 2020-06-25
test: updates on connectObservable, suspense

voliva:
This PR covers the rest of test improvements identified in #20 

For the "it works in suspense" what I did is in `connectObservable` cover the part where it integrates with react (by using exclusively SUSPENSE), and then all the "sugar operators" are tested in isolation.

IMPORTANT -> I made a change to `switchMapSuspended` which I think it makes it a bit smarter. I ran into that when thinking on tests for that operator.

As it's currently in main branch, if a new update comes from the source stream before the inner stream has published anything, that inner stream will get cancelled, a new SUSPENSE will be emitted and a new subscription to the inner stream will be made.

I think it would make sense for `switchMapSuspended` for only emitting one single SUSPENSE while there are no new data. I guess it doesn't hurt for it to emit more than one (or does it?), so I put this feature in a separate commit if we decide we don't want this behaviour and the commit can be dropped. Maybe you want it in another PR?

### DIFF
--- a/test/connectObservable.test.tsx
+++ b/test/connectObservable.test.tsx
@@ -1,14 +1,14 @@
-import { fireEvent, render, screen } from "@testing-library/react"
-import { renderHook, act } from "@testing-library/react-hooks"
-import React, { Suspense, useRef, useEffect } from "react"
-import { BehaviorSubject, concat, defer, from, of, Subject } from "rxjs"
-import { delay, mergeMapTo, scan, take } from "rxjs/operators"
 import {
-  connectObservable,
-  suspend,
-  suspended,
-  switchMapSuspended,
-} from "../src"
+  act as componentAct,
+  fireEvent,
+  render,
+  screen,
+} from "@testing-library/react"
+import { act, renderHook } from "@testing-library/react-hooks"
+import React, { Suspense, useEffect, useRef, FC } from "react"
+import { BehaviorSubject, defer, from, of, Subject } from "rxjs"
+import { delay, scan, startWith, map } from "rxjs/operators"
+import { connectObservable, SUSPENSE } from "../src"
 
 const wait = (ms: number) => new Promise(res => setTimeout(res, ms))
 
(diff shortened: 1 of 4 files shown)


test/connectObservable.test.tsx
```diff
@@ -136,47 +198,52 @@ describe("connectObservable", () => {
...
 
     expect(screen.queryByText("Result 1")).not.toBeNull()
     expect(screen.queryByText("Waiting")).toBeNull()
+  })
 
-    fireEvent.click(screen.getByText(/Next/i))
-
-    expect(screen.queryByText("Result")).toBeNull()
-    expect(screen.queryByText("Waiting")).not.toBeNull()
+  it("keeps in suspense if more than two SUSPENSE are emitted in succesion", async () => {
```
voliva, line 214: > I guess it doesn't hurt for it to emit more than one (or does it?)

I just added this test and it verifies that it doesn't hurt - it doesn't break suspense.

test/connectObservable.test.tsx
```diff
@@ -35,6 +35,49 @@ describe("connectObservable", () => {
...
+    const { result } = renderHook(() => useNumber())
+    expect(result.current).toBe(1)
+
+    act(() => {
+      numberStream.next(2)
+    })
+    expect(result.current).toBe(2)
+  })
+
+  it("allows React to batch synchronous updates", async () => {
```
voliva, line 50: This is the closest I could to try any batching stuff - What this is doing is by testing a very simple case: updates happening in a callback controlled by React (e.g. events) are getting batched.

I wonder if this test actually adds any value. I don't know the specifics of `act`, but if it's calling `unstable_batchUpdates` internally, then the value of this test is basically that updates are being dispatched in the same event loop.

test/connectObservable.test.tsx
```diff
@@ -136,47 +198,52 @@ describe("connectObservable", () => {
...
 
-    await wait(110)
+    render(<TestSuspense />)
 
-    expect(screen.queryByText("Result 2")).not.toBeNull()
+    expect(screen.queryByText("Result 0")).not.toBeNull()
     expect(screen.queryByText("Waiting")).toBeNull()
 
-    fireEvent.click(screen.getByText(/Next/i))
+    await wait(60)
```
josepot, line 236: Let's use subjects rather than `wait`. `wait` ends up generating flaky tests and slowing tests down.

test/operators/switchMapSuspended.test.ts
```diff
@@ -0,0 +1,40 @@
...
+        a: "a",
+      })
+    })
+  })
+
+  it("emits another SUSPENSE when another inner stream activates", () => {
+    scheduler().run(({ expectObservable, cold }) => {
+      const source = cold("-x----x")
+      const inner = cold("  ----a")
+      const expected = "   -s---as---a"
```
josepot, line 30: Let's ensure the switch behavior.

Issue #41, opened 2020-06-28
v0.2.0


josepot:
I think that we've almost at a point where I would feel good releasing `v0.2.0` of this library.

Things that I would like to discuss before `v0.2.0` is released:

- I've got used to the name `distinctShareReplay`, but I hate it. I think that it's a horrible name... I'm actually considering the idea of coining a term for the kind of Observable that it generates (like `BehaviorObservable` or `ReactObservable` or WHATEVER) and then renaming the operator to something like `asReactObservable` or `asBehaviorObservable`.

I'm also starting to wonder whether this operator should be exposed in the first place... But I think that it should. Because we know from experience that if we don't expose it, then most of the users of this library will have to create their own custom `shareReplay`, and the worst part is that it's not even possible to create the custom `shareReplay` that would work best with this library by using any of the overloads of the current RxJS `shareReplay`... So, yeah, I think that it should be exposed, but not with that name... Also, maybe we should only export the "customShareReplay" behavior without the `distinct`? and then we could simply call it `shareLatest`? :thinking: Or maybe we could just name it `shareLatest` and that's it (with the distinct behavior still on it)?

I won't release `v0.2.0` until I'm not happy with this.

There are 2 other things that I would like to happen before we release `v0.2.0`:

1) use docosaurus to generate a site for https://re-rxjs.org or for https://rerxjs.org (I bought both domains). Ideally I would like to have in that site similar to the one that recoil.js has, at least with the following sections:
- Introduction
- Basic Tutorial
- API Reference

2) A blog-post announcing the release of this library.

I've been working on those 2 items, but I will need help... I will create separate issues for those 2. Oh, the docs-site would live under a completely different repo.

voliva:
That's a nice discussion. On distinctShareReplay, I also wondered the same thing. Is it really needed if people can use `connectObservable` or `connectFactoryObservable` and get the "distinctShareReplay"ed observable that can be used to build other streams?

What's the use case where distinctShareReplay would be useful for a consumer?

If we expose it, I like `shareLatest` name, but I wouldn't make it apply `distinctUntilChanged`. From a consumer PoV the useful bit of `shareLatest` would be on how it shares the source stream - distinctUntilChanged can be added if needed.

josepot:
I've been thinking a lot about this and I think that I've finally realized what the issue is here. I will try to summarize it:
- The `distinctUntilChanged` behavior should not be baked into the current `distinctUntilChanged`
[...]

PR #42, opened 2020-06-28
feat(shareLatest): removing distinctShareReplay in favor of shareLatest

josepot:
I'm so so so so so happy about these changes that I can't wait to get them merged! :heart_eyes:

These refactor addresses [what we've discussed](https://github.com/re-rxjs/re-rxjs/issues/41#issuecomment-650778069) on #41 . I think that we finally managed to get this right. Thanks a lot @voliva for the feedback and the help!

FTR: our tests are pretty bad for documenting the behavior of the API and that's something that we should improve ASAP. However, they are great for addressing refactors like this one.


josepot:
And on top of that: have a look at the new bundle-size :heart_eyes: !!! Yes! this is it! This should be the final API for `v0.2.0`, yes! finally!

src/connectObservable.ts
```diff
@@ -13,30 +12,19 @@ import { ConnectorOptions, defaultConnectorOptions } from "./options"
...
 ) {
-  const _options = {
-    ...defaultConnectorOptions,
-    ...options,
-  }
-  const sharedObservable$ = distinctShareReplay(_options.compare)(
-    concat(observable, NEVER),
-  )
-
+  const sharedObservable$ = shareLatest<T>(false)(concat(observable, NEVER))
```
voliva, line 23: with `shareLatest(false)` concat NEVER is redundant, right?

test/operators/share-latest.test.ts
```diff
@@ -11,32 +10,6 @@ const scheduler = () =>
   })
 
 describe("operators/distinctShareReplay", () => {
```
voliva, line 12:
```suggestion
describe("operators/shareLatest", () => {
```

src/connectObservable.ts
josepot: yes, indeed! thanks!

PR #46, opened 2020-07-02
feat: observables from connect* should complete

josepot:
I was thinking about the docs and how to explain why the observables returned from `connectObservable` and `connectFactoryObservable` do not complete and the more I tried to make sense of it, the less sense it made... Then I realized that we were in fact exposing an implementation detail of the underlying "react-observable" and that in many cases that could be very  annoying for the consumers. So, this PR addresses that issue and now I'm finally happy with the API. :smile:


josepot:
FTR: once this is merged I'm going to make a complete re-write of the tests. Having the kind of tests that we got until now it did make some sense while we were constantly making re-factors and API changes, but now the API has reached a point of maturity that I'm confident that we should only write tests against the public API. Also, the tests will now become a lot more descriptive.

src/connectFactoryObservable.ts
```diff
@@ -37,32 +38,29 @@ export function connectFactoryObservable<
...
       sharedObservable$,
       unsubscribeGraceTime,
     )
 
-    const result: [BehaviorObservable<O>, BehaviorObservable<O>] = [
-      sharedObservable$,
+    const result: [Observable<O>, BehaviorObservable<O>] = [
+      takeUntilComplete(sharedObservable$),
       reactObservable$,
     ]
```
voliva, line 65: It's a bit unrelated to this PR, but I always found confusing the fact that the cache holds `[sharedObservable, reactObservable]` and this method returns `[reactHook, sharedObservable]` 😅

src/createInput.ts
```diff
@@ -32,9 +32,7 @@ const createInput_ = <T>(defaultValue: T = empty) => {
     if (defaultValue !== empty) {
       subject.next((current.latest = defaultValue))
     }
-    const source = shareLatest(true, () => cache.delete(key))(
-      subject,
-    ) as Observable<T>
+    const source = shareLatest(subject, () => cache.delete(key))
```
voliva, line 35: All good 👍  
I was thinking - the original had complete:true so maybe it would need takeUntilComplete? => Nope, the subject by definition of `createInput` never completes.

josepot: True story: since it was indifferent whether to use `true` or `false`, I decided to go with `true` because it had less characters :grimacing:  (smaller bunlde-size :see_no_evil:)

src/connectFactoryObservable.ts
josepot: Dah! This is a very good point, I will address that in a separate commit, for sure. Thanks for pointing that out @voliva

PR #47, opened 2020-07-02
tests: only test the public API

josepot:
This is still a WIP, but I will put it here so that we can discuss the tests that we want to add/change, etc.

So, as I already said before: I think that we have reached a point of maturity where it makes sense to make a final investment on the tests. Therefore, I have deleted all the "internal". Surprisingly, after deleting all the tests that were not testing the public API we still got a 99.5% test coverage :smile:. However, that doesn't change the fact that we should improve the quality of our tests so that they are more self-documenting. Also, I want to test things like making sure that certain changes only trigger the minimum number of re-renders, etc.

So, please share in the comments what your thoughts are on how to improve these tests.


josepot:
Ok, there is still room for improvement, for sure... But hey! we got 100% code-coverage through tests that only test the public API, so I think that this is good for now. Objections are very welcome: let me know what tests I should be adding and I will add them, or send a PR with the tests that you think are missing.

voliva:
Sorry I haven't taken a look yet, I'm a bit busy today.

I'll try this evening/night

josepot:
oh, no worries. I didn't mean it that way. Enjoy your weekend! It's just that I wanted those tests in before #48 goes in

PR #48, opened 2020-07-04
connectDispatcher WIP

josepot:
After today's conversation with @bhavesh-desai-scratch I came to realize that we need to provide something to help users create dispatchers and after a TON of thought, this is what I came up with, it could be use like this... Tomorrow I will explain it better because it's passed 4:00 am in BCN. However, @voliva heads up because the hook produced by `connectDispatcher` always returns a stable function and I would love to know if there is a way to extend the "react-hooks/rules-of-hooks" eslint rule so that it can accommodate these callbacks too... probably not, but I don't know, maybe you will find a way :wink:

### DIFF
--- a/src/connectDispatcher.ts
+++ b/src/connectDispatcher.ts
@@ -1,0 +1,51 @@
+import { Observable } from "rxjs"
+import shareLatest, { cache as sharedCache } from "./internal/share-latest"
+import reactEnhancer, { cache as reactCache } from "./internal/react-enhancer"
+import { useEffect } from "react"
+
+const dispatchers = new WeakMap<Observable<any>, any>()
+
+const getReactObservable$ = <T>(input$: Observable<T>) => {
+  if (!sharedCache.has(input$)) {
+    input$ = shareLatest(input$)
+  }
+  if (!reactCache.has(input$)) {
+    reactEnhancer(input$, 200)
+  }
+  return reactCache.get(input$)!
+}
+
+export function connectDispatcher<
+  A extends (number | string | boolean | null)[],
+  Input,
+  Output
+>(
+  getObservable: ((...args: A) => Observable<Output>) | Observable<Output>,
+  dispatcher: (input: Input, ...args: A) => void,
+) {
+  const getHookObservable$ = (...args: A) =>
+    getReactObservable$(
+      typeof getObservable === "function"
+        ? getObservable(...args)
+        : getObservable,
+    )
+
+  const getDispather = (...args: A) => {
+    const observable$ = getHookObservable$(...args)
+    let result: (input: Input) => void = dispatchers.get(observable$)
+    if (result) return result
+    result = (input: Input) => {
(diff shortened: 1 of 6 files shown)


src/connectDispatcher.ts
```diff
@@ -0,0 +1,51 @@
...
+
+const getReactObservable$ = <T>(input$: Observable<T>) => {
+  if (!sharedCache.has(input$)) {
+    input$ = shareLatest(input$)
+  }
+  if (!reactCache.has(input$)) {
+    reactEnhancer(input$, 200)
+  }
+  return reactCache.get(input$)!
+}
```
voliva, line 16: I'm a bit confused: Why does it need to be `shareLatest`-ed and `reactEnhancer`-ed? As I'm reading how this is being used and probably would be used by users, I don't think this should need to apply these methods:
* `shareLatest` shouldn't be needed because the `useEffect` down below will use the stream with shareLatest on it, but only only applies to that internal useEffect.
* `reactEnhancer` shouldn't be needed because it's ignoring every message, so react doesn't really play a role in the behaviour of that stream.

voliva:
> @voliva heads up because the hook produced by connectDispatcher always returns a stable function and I would love to know if there is a way to extend the "react-hooks/rules-of-hooks" eslint rule so that it can accommodate these callbacks too... probably not, but I don't know, maybe you will find a way 😉

It would be great, but unfortunately I couldn't think of anything elegant - The source of rules-of-hooks that checks for exhaustive dependencies is [this one](https://github.com/facebook/react/blob/master/packages/eslint-plugin-react-hooks/src/ExhaustiveDeps.js#L369) and it's got `useRef`, `useState` and `useReducer` hardcoded in, and there's no way of extending it.

It could probably be enhanced though - The hacky idea I have in mind is:
* In our eslint-plugin add a configuration that disables the "rules-of-hooks/exhaustive-deps" rule
* In our eslint-plugin add a rule "react-rxjs/exhaustive-deps" that imports the original react rule from `eslint-plugin-react-hooks` and proxies it through.
* Check whenever the react rule reports an error on a dependency list that is only missing our target, and drop that error.

Problem is... this is not really composable. What if there's another library that does the same? Not only that, but we'd have to see how to deal with auto-fixer for that rule. I don't think we should do it this way. 

Another posible front is to try and push react to make that configurable. They have done it for hooks which [should have their dependencies checked](https://www.npmjs.com/package/eslint-plugin-react-hooks#advanced-configuration) (as if you'd like to implement a custom useEffect), but I still need to see if it's posible from our plugin to append a configuration of another plugin.

josepot:
yep... Screw this! I should sleep more. Thanks @voliva

josepot:
Although, there is one thing from this set of changes that I want to keep: `createInput` should be stateless. No initial state, no setter/reducer overload... If you want a counter, put the "reducer" in the stream and then emit the deltas or "INC"/"DEC" signals or something. That I'm going to change, for sure.

PR #50, opened 2020-07-04
feat(createSubjectsFactory): deprecated createInput for createSubjectsFactory

josepot:
~Also, I don't like this name. Let's try to come up with a better name... Wouldn't something like `createFactory` be a better name? :thinking:~

I ended up deprecating `createInput` in favor of this `createSubjectsFactory`.

Please @voliva let me know what your thoughts are no this.

FTR: I want to ship this utility. However, it almost feels like this utility kinda belongs into another package on top of `react-rxjs`. Something like `react-rxjs-utils` or something like that :thinking:


test/createInput.test.ts
```diff
@@ -45,7 +38,7 @@ describe("createInput", () => {
   })
 
   test("it replays the latest value to new subscriptions", () => {
```
voliva, line 40: Food for thought: should it?

josepot: Good point! Now that it's stateless, it shouldn't, no!

josepot:
I'm still not happy with the name, though... What about just: `subjectsFactory` or `newSubjectsFactory`?

voliva:
I'd just drop the plural `s` within `subjects`. Any combination of  `create/new/null` seems fine to me :3

As for the question to ship it in another package, I also think this is a rxjs helper that we've find useful when working with react-rxjs but it's not directly related to it. However as a user, it might become a bit confusing to have many packages with some utils in one and some utils (e.g. `shareLatest`) in another.

josepot:
Hi @voliva could you please re-review it? I just changed the name to `subjectFactory`, I like it better than the previous ones but I'm still not completely satisfied with this name... Hopefully we will come up with the "right name" before the release of the stable version, or maybe we will just get used to this name :shrug: :smiley:

Should I merge?

README.md
```diff
@@ -177,53 +177,40 @@ const story$ = selectedStoryId$.pipe(
...
-    {counter}
-    <button onClick={() => setCounter$(id, x => x + 1)} />+</button
-  )
-}
-```
+const getCounterActions$ = subjectFactory<string, 'INC' | 'DEC'>()
 
-or:
+const onInc = (id: string) => getCounterActions$(id).push('INC')
+const onDec = (id: string) => getCounterActions$(id).push('DEC')
```
voliva, line 186:
```suggestion
const onInc = (id: string) => getCounterActions$(id).next('INC')
const onDec = (id: string) => getCounterActions$(id).next('DEC')
```

josepot: :facepalm: Thanks!

PR #53, opened 2020-07-08
test(connectObservable): add test combining ErrorBoundary and Suspense

voliva:
I've found this issue where if a stream throws an error before emitting any value, the error can't be caught in an ErrorBoundary.

I've made this test that describes the issue and fails with the current version. A codesandbox can be found [here](https://codesandbox.io/s/react-rxjs-playground-dm3tl?file=/src/App.tsx)

For reference, react uses this [sandbox](https://codesandbox.io/s/reactsuspenseerror-b5qu0?file=/src/fakeApi.js) to show how the error is caught also in data fetching. I had a quick look and I think that they have it working because  the component throws an error if the promise rejected.

### DIFF
--- a/src/internal/react-enhancer.ts
+++ b/src/internal/react-enhancer.ts
@@ -1,5 +1,5 @@
 import { Observable, of, Subscription, Subject, race } from "rxjs"
-import { delay, takeUntil, take, filter, tap } from "rxjs/operators"
+import { delay, takeUntil, take, filter, tap, catchError } from "rxjs/operators"
 import { SUSPENSE } from "../SUSPENSE"
 import { BehaviorObservable } from "./BehaviorObservable"
 import { EMPTY_VALUE } from "./empty-value"
@@ -57,7 +57,12 @@ const reactEnhancer = <T>(
   }) as BehaviorObservable<T>
 
   let promise: any
+  let hasError = false
+  let error: unknown = undefined
   const getValue = () => {
+    if (hasError) {
+      throw error
+    }
     try {
       return (source$ as BehaviorObservable<T>).getValue()
     } catch (e) {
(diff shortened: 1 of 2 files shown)


src/internal/react-enhancer.ts
```diff
@@ -57,15 +57,23 @@ const reactEnhancer = <T>(
   }) as BehaviorObservable<T>
 
   let promise: any
+  let hasError = false
+  let error: unknown = undefined
   const getValue = () => {
+    if (hasError) {
+      throw error
+    }
```
voliva, line 65: I compared with the way that this is solved in React's example and within their `read` (our `getValue()` equivalent) they either return the value, throw the promise, or throw the error.

src/internal/react-enhancer.ts
```diff
@@ -57,15 +57,23 @@ const reactEnhancer = <T>(
...
       if (promise) throw promise
 
       if (!IS_SSR && e !== SUSPENSE) {
         source$
-          .pipe(takeUntil(race(onSubscribe, of(true).pipe(delay(60000)))))
+          .pipe(
+            takeUntil(race(onSubscribe, of(true).pipe(delay(60000)))),
+            catchError(() => of(true)),
+          )
           .subscribe()
```
voliva, line 77: This was needed because otherwise this subscription would receive an error without handling it - Raising an uncaught exception that jest would pick up as `error` even if (`expect(true).toBe(true)`).

test/connectObservable.test.tsx
```diff
@@ -266,6 +266,39 @@ describe("connectObservable", () => {
...
+      </TestErrorBoundary>,
+    )
+
+    componentAct(() => {
+      errStream.error("controlled error")
+    })
+
+    await componentAct(async () => {
+      await wait(0)
+    })
```
voliva, line 294: Ewww.

I couldn't find any doc on this, but by experimenting, it seems that when a thrown promise fails in Suspense:
* It creates a new component instance. Yep, the reducer `init` function gets called, and if you have a `setState(() => Math.random())` you'll get another value.
* It retries rendering the component a couple of times.

It looks like the single `componentAct` can't pick this up. Chaining two toghether does fix it.

Do you smell that smelly smell?

voliva:
Closing this in favour of #54

PR #54, opened 2020-07-09
fix(error-handling): fix error-handling issues with suspense

josepot:
This was fun! :sweat_smile:

This PR finishes the work that @voliva started on #53 
Thanks a lot @voliva for reporting this issue and for all the details that you've provided in #53, we should go over all this stuff at some point.

### DIFF
--- a/src/internal/useObservable.ts
+++ b/src/internal/useObservable.ts
@@ -2,22 +2,22 @@ import { useEffect, useReducer } from "react"
 import { BehaviorObservable } from "./BehaviorObservable"
 import { SUSPENSE } from "../SUSPENSE"
 
+const ERROR: "e" = "e"
+const VALUE: "v" = "v"
+const SUSP: "s" = "s"
+type Action = "e" | "v" | "s"
+
 const reducer = (
-  _: any,
-  { type, payload }: { type: "next" | "error"; payload: any },
+  _: { type: Action; payload: any },
+  action: { type: Action; payload: any },
 ) => {
-  if (type === "error") {
-    throw payload
+  if (action.type === ERROR) {
+    throw action.payload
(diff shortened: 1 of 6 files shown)


src/internal/useObservable.ts
```diff
@@ -51,5 +52,8 @@ export const useObservable = <O>(
     return () => subscription.unsubscribe()
   }, [source$])
 
-  return state !== (SUSPENSE as any) ? (state as any) : source$.getValue()
+  if (state === SUSPENSE) {
+    throw source$.getValue()
+  }
+  return state
```
voliva, line 58: Can't this have some sort of race condition where `source$.getValue()` would return a value when `state === SUSPENSE`?

test/connectFactoryObservable.test.tsx
```diff
@@ -223,6 +231,125 @@ describe("connectFactoryObservable", () => {
...
+      )
+      unmount()
+    })
+
+    it("allows async errors to be caught in error boundaries with suspense", async () => {
+      const errStream = new Observable(observer => {
+        setTimeout(() => {
+          observer.error("controlled error")
+        }, 100)
+      })
```
voliva, line 267: If we use a subject instead, we could save some time when running these tests - We can try to see if that would work.

src/internal/useObservable.ts
josepot: Mmmmm _maybe_, as in we dispatch `SUSPENSE` from the `useEffect`, but since it's async, then by the time that we return, then the promise is gone... yeah, that's a problem... Thanks for pointing that out!

josepot: Done!

PR #55, opened 2020-07-12
feat(connectFactoryObservable): accepts an object as the key

josepot:
This is something that was already proposed and discussed on #9 and I ultimately [decided against this addition](https://github.com/re-rxjs/react-rxjs/pull/9#issuecomment-643824404) unless we could justify it with non contrived use-cases.

A this point we have already come-up with a couple of those use-cases and it's pretty clear to me that despite it's potential misusage, this addition can be very useful.

### DIFF
--- a/src/connectFactoryObservable.test.tsx
+++ b/src/connectFactoryObservable.test.tsx
@@ -9,7 +9,7 @@ import {
   Subject,
 } from "rxjs"
 import { renderHook, act as actHook } from "@testing-library/react-hooks"
-import { switchMap, delay } from "rxjs/operators"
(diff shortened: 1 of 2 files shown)


josepot:
I'm still on the fence on this one... I will wait a couple more days before I merge this.

josepot:
I'm closing this once again. The reason being that for the use-cases that I found there was a "better" way to model the streams which didn't this.

Also, if we ever want to add this feature, then we should make sure that it's also possible to combine primitives with objects... I have an idea of how that would look like, but let's not get ahead of ourselves, because I'm still not sure that this is actually needed.

PR #56, opened 2020-07-14
fix(connectObservable): fix updates when dealing with many component

voliva:


src/internal/react-enhancer.ts
```diff
@@ -11,17 +11,18 @@ const reactEnhancer = <T>(
...
+  let initialSyncValue = EMPTY_VALUE
   const result = new Observable<T>(subscriber => {
     let isActive = true
     const subscription = source$.subscribe({
       next(value) {
         if (
           isActive &&
           value !== (COMPLETE as any) &&
-          !Object.is(latestValue, value)
+          !Object.is(initialSyncValue, value)
```
voliva, line 22: This was the cause. `latestValue` is shared among all subscribers of the same observable, so only the first one would get the value.

This check was added in #54, and IRC it was done for the batch updates test (at least, it's the test that fails if I remove this check)

PR #60, opened 2020-07-17
Fix a few typos and invalid code in README.md

mattmischuk:
Found a couple syntax errors in TSX examples. I also replaced two instances of `<div />`s being used instead of `<p />`. Happy to revert that change if you prefer `<div />`.


josepot:
Thanks a lot for this @mattmischuk ! Much appreciated!

@all-contributors please add @mattmischuk for docs

PR #63, opened 2020-07-20
docs(monorepo): update README files

josepot:
Having the API in these READMEs is something temporary while we finish the react-rxjs.org docs site. Once that's done these readmes will just redirect the users to the relevant sections of react-rxjs.org.

### DIFF
--- a/packages/core/README.md
+++ b/packages/core/README.md
@@ -1,0 +1,127 @@
+# @react-rxjs/core
+
+## Installation
+    npm install @react-rxjs/core
+
+## API
+
+### connectObservable
+```ts
+const [useCounter, sharedCounter$] = connectObservable(
+  clicks$.pipe(
+    scan(prev => prev + 1, 0),
+    startWith(0),
+  )
+)
+```
+Accepts: An Observable.
+
+Returns `[1, 2]`
+
+1. A React Hook that yields the latest emitted value of the observable. If the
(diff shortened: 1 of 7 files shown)


packages/dom/README.md
```diff
@@ -0,0 +1,24 @@
+# @react-rxjj/dom
```
voliva, line 1:
```suggestion
# @react-rxjs/dom
```

packages/core/README.md
```diff
@@ -0,0 +1,127 @@
+# @react-rxjj/core
```
voliva, line 1:
```suggestion
# @react-rxjs/core
```

packages/utils/README.md
```diff
@@ -0,0 +1,119 @@
+# @react-rxjj/utils
```
voliva, line 1:
```suggestion
# @react-rxjs/utils
```

packages/utils/src/groupInMap.ts
```diff
@@ -50,3 +53,46 @@ export const groupInMap = <T, K, V>(
     }),
   )
 }
+
+const votesByKey$ = new Subject<{key: string}>()
```
voliva, line 57: ^^^ I think you forgot to remove the example from the source

josepot: :facepalm:

PR #64, opened 2020-07-21
feat(core): bind

josepot:
Shall we `bind`?


josepot:
shut! It turns out that [the build failed](https://travis-ci.org/github/re-rxjs/react-rxjs/builds/710304297) and we are getting the green button? :thinking: That's not good...

josepot:
Ok... Let's `bind`, I guess... :grimacing: I'm a bit nervous about this TBH, but let's do it! :rocket:

PR #73, opened 2020-07-22
chore: refactor avoid object subscriptions

josepot:
Size matters! :joy: (my guess is that this should decrease the bundle-size a tiny-tiny bit :joy:)


josepot:
Look at that! `-25B, -0.5%` :joy: :tada: One byte at the time :slightly_smiling_face:

PR #75, opened 2020-07-28
feat(utils): add selfDependant creation operator

josepot:


josepot:
Closing this because I realized that this is actually fluffy AF... I initially thought that the `observeOn(asapScheduler)` was necessary, but thanks to @voliva I've realized that's not the case. Therefore, this is just creating a Subject and `tap`ing it, so I now think that this doesn't add any significant value

josepot:
I've ended up changing my mind about this. Yes, what this operator does is trivial... However:

- This is a more declarative way of defining cyclical/recursive Observables. In the moment of its definition you already tell what that is about.

- This is a pattern that's difficult to get right and that it's worth promoting. For instance: trying to use a `withLatestFrom` with a `defer` returning the self-dependant observable is a tempting approach, but it's wrong because it generates a subscription that can't be teared down.

- In the future, using this operator can make the integration with the dev-tools (rxjs-traces) easier (on a development build).

@voliva shall we?

voliva (approved):
Yep! I appreciate that 👍

PR #78, opened 2020-08-15
feat(bind): factory with nested keys

josepot:
A WIP to explore @voliva suggestion's  on [this comment](https://github.com/re-rxjs/react-rxjs/pull/77#issuecomment-674357623) of #77


josepot:
After the offline conversation that we've had, I had an idea to support optional args (and undefined values). I've put that change in a [separate commit](https://github.com/re-rxjs/react-rxjs/pull/78/commits/a6e90d670f299d446b91959afc3f699fa7ab9804) so that it's easier to appreciate what the trick is about 😄

packages/core/src/bind/connectFactoryObservable.ts
```diff
@@ -6,6 +6,52 @@ import { useObservable } from "../internal/useObservable"
 import { SUSPENSE } from "../SUSPENSE"
 import { takeUntilComplete } from "../internal/take-until-complete"
 
+class NestedMap<K extends [], V extends Object> {
```
voliva, line 9: I'd move this in a separate file... I think it's a bit noisy here

packages/core/src/bind/connectFactoryObservable.ts
```diff
@@ -28,30 +74,30 @@ import { takeUntilComplete } from "../internal/take-until-complete"
...
   const getSharedObservables$ = (
-    ...input: A
+    input: A,
   ): [Observable<O>, BehaviorObservable<O>] => {
-    const key = JSON.stringify(input)
-    const cachedVal = cache.get(key)
+    for (let i = input.length - 1; input[i] === undefined && i > -1; i--) {
+      input.splice(-1)
+    }
+    const keys = ([input.length, ...input] as any) as A
```
voliva, line 92: This reminded me of something exciting: The TS release for this month (4.0) will include [variadic tuple types](https://github.com/microsoft/TypeScript/pull/39094) - Which will allow us to represent this type as `[number, ...A]`.

packages/core/src/bind/index.ts
```diff
@@ -46,12 +46,18 @@ export function bind<T>(
  * subscription, then the hook will leverage React Suspense while it's waiting
  * for the first value.
  */
-export function bind<A extends (number | string | boolean | null)[], O>(
+export function bind<
+  A extends (number | string | boolean | null | Object | Symbol | undefined)[],
+  O
+>(
```
voliva, line 52: Do we need to specific what types it supports? Could it be `unknown[]` or `any[]` instead?

packages/core/src/bind/connectFactoryObservable.ts
josepot: I decided to move it below (in the same file). The reason being that `NestedMap` is not generic enough to be consumed from anywhere else, it's very tightly coupled to this functionality. So, I decided to to move it below.

PR #79, opened 2020-08-16
feat(utils): add split operator

josepot:
~I still have to add its tests, but after thinking a lot about this, I think that we would really benefit from this operator and I wanted to share what I got so far.~

This operator is very similar to the official `groupBy` operator of `rxjs`. However, it has a couple important differences that IMO make its usage more ergonomic when working with the `react-rxjs` bindings. The main differences are:
- There are only two arguments, and the second one is optional.
- The second argument is what IMO makes things more ergonomic, it's a `streamSelector`, which receives the original grouped observable and returns another observable which will end up being the resulting grouped observable for that key.
- The grouped observables replay their latest emitted value.
- There is no need for a `durationSelector` because the duration is determined by the duration of the grouped observable.

For the tests I've borrowed and adapted a bunch of tests from the official `rxjs` repo, I didn't cherry-pick the ones that passed, I just picked the ones that I thought were more relevant to `split`. I still want to add a few more, but for now this should do. @voliva if you think of some tests that should be added before this gets merged, please let me know it.

We should probably add some tests for the "replayable" behaviour of the inner observables 🤔


packages/utils/src/split.ts
```diff
@@ -0,0 +1,57 @@
...
+          const res = (streamSelector
+            ? streamSelector(subject, key).pipe(shareReplay(1))
+            : subject.asObservable()) as GroupedObservable<K, R>
+
+          res.key = key
+          res.subscribe({
+            complete() {
+              groups.delete(key)
+            },
+          })
```
voliva, line 42: This subscription will need cleaning up in the following scenario:

```js
cold('1-2--3-4').pipe(
  split(v => v % 2),
  take(2)
).subscribe()
```

As there's nothing completing the inner subject.

I was thinking this is only relevant when you don't pass in a `streamSelector` parameter, because if you do then you have control when to stop it. But I can't think of an elegant way to solve this same example using streamSelector.... `takeUntil(what?)`: At that point we don't know the logic of the consumer (when will it unsubscribe?).

josepot: Good catch! I've addressed that and I've added [a test for this case](https://github.com/re-rxjs/react-rxjs/pull/79/files#diff-0c49ff00c654149cec7eaa7af1e52bdcR24-R37).

josepot:
> I'd like to add a test to check it handles synchronous bursts on subscribe - I reckon it should work as it is now, but I think it's a nice case to have.

Sorry, but I don't quite understand what's that you want to test here... Could you please provide an example?

voliva:
Yep:

```js
  it("should handle synchronous values", () => {
    scheduler().run(({ expectObservable, cold }) => {
      const e1 = from(["1", "2", "3", "4", "5"])
      const expected = "(xy|)"
      const x = cold("(135|)")
      const y = cold("(24|)")
      const expectedValues = { x: x, y: y }
      const source = e1.pipe(split((val: string) => parseInt(val) % 2))
      expectObservable(source).toBe(expected, expectedValues)
    })
  })
```

As I said, I don't see why this shouldn't work, but it's a case that often fails when managing inner subscriptions.

josepot:
> Yep:
> 
> ```js
>   it("should handle synchronous values", () => {
>     scheduler().run(({ expectObservable, cold }) => {
>       const e1 = from(["1", "2", "3", "4", "5"])
>       const expected = "(xy|)"
>       const x = cold("(135|)")
>       const y = cold("(24|)")
>       const expectedValues = { x: x, y: y }
>       const source = e1.pipe(split((val: string) => parseInt(val) % 2))
>       expectObservable(source).toBe(expected, expectedValues)
>     })
>   })
> ```
> 
> As I said, I don't see why this shouldn't work, but it's a case that often fails when managing inner subscriptions.

Yep, now I get it! I agree it's a common issue when working with inner subscriptions, so I've added the test. Should I merge this?

voliva (approved):
Yep 💯

PR #80, opened 2020-08-20
feat(utils): deprecate groupInMap in favour of collectValues

josepot:
This is a breaking change, but I really think that this is the way to go, specially now that `split` has been added into `utils`.

Right after we added `groupInMap` I felt like the operator was doing a bit too much, because in many cases it's very convenient to have access to the underlying Observable of GroupedObservables, and this is what this PR is about.

Notice how in the tests the only thing that has changed is that before we were doing:

```ts
groupInMap(
  (value) => value.key,
  (value$) => value$.pipe(map((value) => value.quantity))
)
```

And now we can accomplish the same by doing:

```ts
split(
  (value) => value.key,
  (value$) => value$.pipe(map((value) => value.quantity))
),
groupInMap()
```


packages/utils/src/groupInMap.ts
```diff
@@ -18,35 +10,23 @@ const DELETE = Symbol("DELETE")
...
-    defer(() => {
-      res.clear()
-      return of(res)
-    }),
+    scan((acc, [key, value]) => {
+      if (value !== DELETE) return acc.set(key, value)
+      acc.delete(key)
+      return acc
+    }, res),
+    debounceTime(0, asapScheduler),
```
voliva, line 30: Why is this needed? Wouldn't it be better if it was handled by the consumer?

packages/utils/src/groupInMap.ts
```diff
@@ -18,35 +10,23 @@ const DELETE = Symbol("DELETE")
...
  * @param projection Projection function for each group.
  */
-export const groupInMap = <T, K, V>(
-  keyGetter: (x: T) => K,
-  projection: (x: GroupedObservable<K, T>) => Observable<V>,
-) => (source$: Observable<T>): Observable<Map<K, V>> => {
+export const groupInMap = <K, V>() => (
+  source$: Observable<GroupedObservable<K, V>>,
+): Observable<Map<K, V>> => {
   const res = new Map<K, V>()
```
voliva, line 16: I think different subscriptions will use the same reference of this map... Maybe the scan overload you made with a lazy initial value can come in handy?

voliva:
I think this is a great change. And the name `groupInMap` still makes sense in this new usage 👍

packages/utils/src/groupInMap.ts
josepot: Very good point, thanks for pointing that out!

packages/utils/src/groupInMap.ts
josepot: I don't know... I've been thinking about this and I'm on the fence.

The thing is that there will be many instances when a single event will trigger many sync changes for different inner-streams. The most obvious one case is when the source stream completes and then all the inner-streams complete one by one, therefore triggering many different deletions until Map is empty.

Of course, that particular case we can handle it in this operator, so that only one empty Map gets emitted when the source completes. But what about the other cases when an event will trigger many different sync changes? Do we really want to emit all those synchronously? Isn't it too noisy?

You are right, though. If we handle the case of the source stream completion, then we can let the consumer do this debouncing themselves... It's just that I can't think of a single case where I would want to get all that noise when what we are emitting is a Map, that's why I think that it makes sense to handle this on operator itself.

Can come up with an example where doing this would be problematic for the consumer?

josepot: Actually, NVM... I can think of some cases where this can be pretty annoying...

PR #103, opened 2020-09-01
chore: Add Victor as a co-author

josepot:
I started this thing, yep... but I really think of @voliva as a co-author of this library. His insights, his feedback, his proposals, all the contributions that his made to the API, etc, etc. I don't think that it's fair that my name is the only one on the list of authors.

So, @voliva can I add you to the list of "authors"?


rikoe (approved):

voliva (approved):
Thanks a lot for this, and your hard work!

PR #107, opened 2020-09-04
0.3.0

josepot:
This is my proposal for v0.3.0.

Breaking changes:
- Removed the `graceTime` option from `bind`, `useSubsribe` and `Subscribe`
- Subscribe: won't render `children` until after starting the subscription. I've put @voliva as a co-author of the commit because this was his idea and he also sent me a snippet with the code.

The rest of the changes are different kinds of perf-optimizations.

I'm really happy with this new API 😍 . Thanks a lot @voliva for your invaluable help on this.

### DIFF
--- a/packages/core/src/Subscribe.tsx
+++ b/packages/core/src/Subscribe.tsx
@@ -7,15 +7,12 @@ import { Observable } from "rxjs"
  * the component mounts and it unsubscribes when the component unmounts.
  *
  * @param source$ Source observable that the Component will subscribe to.
- * @param graceTime (= 200): Amount of time in ms that the Component should wait
- * before unsubscribing from the source observable after it unmounts.
  *
  * @remarks This Component doesn't trigger any updates.
  */
 export const Subscribe: React.FC<{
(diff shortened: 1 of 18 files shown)


packages/core/src/Subscribe.tsx
```diff
@@ -1,21 +1,24 @@
...
  */
 export const Subscribe: React.FC<{
   source$: Observable<any>
-  graceTime?: number
-}> = ({ source$, graceTime, children }) => {
-  useSubscribe(source$, graceTime)
-  return <>{children}</>
+  fallback?: null | JSX.Element
+}> = ({ source$, children, fallback }) => {
+  const [mounted, setMounted] = useState(0)
```
voliva, line 17: LOL - Nice safe on those 3 extra bytes 😄

voliva (approved):

josepot:
Cool! I will merge this, but i will publish at least one pre-release before I publish v0.3.0, let's try to test the pre-release as much as we can in the next day or two.

PR #108, opened 2020-09-06
perf(core): use useLayoutEffect

josepot:
This is a change that I wanted to add to v0.3.0, but I forgot... Good think that we haven't released it yet! :sweat_smile:


voliva (approved):
👌

PR #112, opened 2020-09-08
fix(utils-split): dismiss inner observable on error

josepot:
ups!

### DIFF
--- a/packages/utils/src/split.test.ts
+++ b/packages/utils/src/split.test.ts
@@ -58,7 +58,7 @@ describe("split", () => {
       )
   })
 
-  it("should group values while the inner stream is active", () => {
+  it("should group values while the inner stream does not complete", () => {
     const expectedGroups = [
       { key: 1, values: [1, 3] },
       { key: 0, values: [2, 4] },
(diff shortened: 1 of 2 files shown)


rikoe (approved):
Is this use case covered by a test?

josepot:
> Is this use case covered by a test?

Of course, our code-coverage is 100%! that means that everything is covered by tests, right?

Just kidding! I've been trying to write a test for this, but I'm having a very hard time coming up with a decent one, but you are right, I won't merge it until I can write that test. Just to be clear, the test should not pass on the current `main` branch and it should pass after these changes.

Thanks @rikoe !

josepot:
@all-contributors please add @rikoe for review

voliva (approved):

rikoe (approved):

josepot:
The first commit is a test that doesn't pass on main ([I made sure that the CI ran on that commit and that the test fails](https://github.com/re-rxjs/react-rxjs/runs/1095688098)) and the second commit is the one that fixes the issue.

PR #118, opened 2020-09-12
Revert "perf(core): use useLayoutEffect"

josepot:
This reverts commit 973669553b98befc4256a3979870de59b4d58950.

I was preparing a PR to `will-this-react-global-state-work-in-concurrent-mode` to update the name of the library and to use the latest version, and as I was doing that I realized that the test `ability to interrupt render` was not passing. It turns out that's because of the theoretical perf improvement of using `useLayoutEffect` rather than using `useEffect`. So, after having thought about this quite a bit, I'm quite positive that we should revert this commit.


josepot:
I'm merging this and publishing a patch update.

PR #120, opened 2020-09-14
fix(core): factory bind multiple (un/re)subscriptions

josepot:
As @voliva pointed out to me offline, the current behaviour of the shared-observable returned from the "factory bind" is problematic, because a consumer could want to resubscribe to it after all its initial subscriptions have finished. Currently this would trigger an error and that's not the intended behaviour.

This PR addresses the behaviour of what it should occur when a new subscription happens on a shared-observable that was previously disposed. There are 2 possible scenarios here:

1) When the new subscription happens there is no instance on the cache for that given key: in this case a new instance will be added and all subsequent subscriptions will consume that instance until it's no longer needed.

2) There is already a new instance on the cache: in this case the new subscription will be wired to that one, so that we continue sharing the latest one.

There is a new test that testes all these behaviours. The test should probably be improved and broken down into 2-3 different tests that better document this behaviour.

I think that we can do that in a later moment, right now my priority is to push a patch with this fix.

### DIFF
--- a/packages/core/src/bind/connectFactoryObservable.test.tsx
+++ b/packages/core/src/bind/connectFactoryObservable.test.tsx
@@ -9,7 +9,7 @@ import {
   Subject,
 } from "rxjs"
 import { renderHook, act as actHook } from "@testing-library/react-hooks"
-import { switchMap, delay } from "rxjs/operators"
+import { switchMap, delay, take } from "rxjs/operators"
 import { FC, Suspense, useState } from "react"
 import React from "react"
 import {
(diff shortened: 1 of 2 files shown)


voliva (approved):
should we create a TODO issue for breaking down this test into smaller ones?

josepot:
> should we create a TODO issue for breaking down this test into smaller ones?

At the end I decided to address this into this PR. Please let me know if you think that this tests are good enough. Thanks!

voliva (approved):
thank you so much! 🚀

PR #127, opened 2020-09-22
utils(createListener): add createListener util

### LINKED ISSUE #126: new util: newInput
Admittedly, creating input streams is a lot of boilerplate. I remember that @rikoe mentioned it once and I agree: it's quite tedious. So, I'm thinking about adding the following util:

```ts
const newInput = <A extends unknown[], T>(                  
  mapper: (...args: A) => T,                                   
): [Observable<T>, (...args: A) => void] => {                  
  const subject = new Subject<T>()                             
  const onEvent = (...args: A) => subject.next(mapper(...args))
  return [subject.asObservable(), onEvent]                     
}                                                              
```

either that, or:

```ts
const newInput = <A extends unknown[], T>(                  
  mapper: (...args: A) => T,                                   
): [Subject<T>, (...args: A) => void] => {                  
  const subject = new Subject<T>()                             
  const onEvent = (...args: A) => subject.next(mapper(...args))
  return [subject, onEvent]                     
}                                                              
```

Should we add it? And if we do, which one do you prefer: the one that exposes the `Subject` or the one that just exposes an `Observable`?

An example:

before:
```ts
const add$ = new Subject<string>()                                
const onAdd = (text: string) => text && userAdd$.next(text)

const edit$ = new Subject<{ id: number; text: string }>()
const onEdit = (id: number, text: string) => userEdit$.next({ id, text })                               
```
after:
```ts
const [add$, onAdd] = newInput((text: string) => text)
const [edit$, onEdit] = newInput((id: number, text: string) => ({ id, text }))
```

josepot:
Closes #126 

After an "offline" conversation with @rikoe he convinced me that `createListener` -despite being a bit verbose- it's a better name than `newInput` or than `createInput`, and I think that he is right. So, here we go. Are we all happy with the name and with the signature?

### DIFF
--- a/packages/utils/src/createListener.spec.ts
+++ b/packages/utils/src/createListener.spec.ts
@@ -1,0 +1,27 @@
+import { createListener } from "./"
+
+describe("createListener", () => {
+  it('receives an "event creator" and it returns a tuple with an observable and its corresponding event-emitter', () => {
+    const [fooBar$, onFooBar] = createListener((foo: number, bar: string) => ({
+      foo,
+      bar,
+    }))
+    let receivedValue
+    fooBar$.subscribe((val) => {
+      receivedValue = val
+    })
+    expect(receivedValue).toBe(undefined)
+    onFooBar(0, "1")
+    expect(receivedValue).toEqual({ foo: 0, bar: "1" })
+  })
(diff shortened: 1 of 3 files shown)


rikoe (approved):
🎉

josepot:
The CI is failing... guess why?

image

Sometimes `unstable_batchedUpdates` batches more than we expected :sob:

PR #136, opened 2020-10-09
fix(core): factory react observable

josepot:


packages/core/src/bind/connectFactoryObservable.ts
```diff
@@ -53,17 +56,21 @@ export default function connectFactoryObservable<A extends [], O>(
...
+      const source$ = inCache
+        ? inCache[0] === publicShared$
+          ? sharedObservable$
+          : inCache[0]
+        : getSharedObservables$(input)[0]
+
+      publicShared$.getValue = source$.getValue
+
+      return source$.subscribe(subscriber)
+    }) as BehaviorObservable<O>
```
voliva, line 70: Something calling `.getValue` before no subscription happens will throw an error. Would it be safer to pre-initialize getValue to ~~`throw EMPTY_VALUE`~~ `sharedObservable$.getValue`?

voliva (commented):
Awesome finding. Would it be posible to have some tests that reproduce the fixed issue?

packages/core/src/bind/connectFactoryObservable.ts
josepot: > Something calling .getValue before no subscription happens will throw an error.

Yep, that is intentional.

> Would it be safer to pre-initialize getValue to throw EMPTY_VALUE sharedObservable$.getValue?

With the current implementation as long as it throws, it doesn't matter what it throws. It would be a lot more clear, for sure, but it wouldn't make a difference.

josepot:
> Would it be posible to have some tests that reproduce the fixed issue?

I'm not sure how to do that without testing an implementation detail... I'm open to suggestions, though. :sweat_smile:

PR #137, opened 2020-10-10
Removing graceTime and adding lots of perf improvements

josepot:
The `graceTime` is a dangerous thing. No one should rely on it, it's an implementation detail. So, I've decided to remove it because it's not as useful as I initial thought it was. Also, it adds lots of unnecessary complexity.

There are 3 different commits in this PR:

1. Gets rid of the graceTimesecond
2. Simplifies *a lot* both: the `reactEnhancer` and `useObservable`
3. Removes `reactEnhancer` :tada:

### DIFF
--- a/packages/core/src/bind/connectFactoryObservable.ts
+++ b/packages/core/src/bind/connectFactoryObservable.ts
@@ -80,7 +80,7 @@ export default function connectFactoryObservable<A extends [], O>(
   return [
     (...input: A) => {
       const [source$, getValue] = getSharedObservables$(input)
-      return useObservable(source$, getValue)
+      return useObservable(source$, getValue, input)
     },
     (...input: A) => getSharedObservables$(input)[0],
   ]
--- a/packages/core/src/bind/connectFactoryObservable.ts
+++ b/packages/core/src/bind/connectFactoryObservable.ts
@@ -30,14 +30,11 @@ export default function connectFactoryObservable<A extends [], O>(
   (...args: A) => Exclude<O, typeof SUSPENSE>,
   (...args: A) => Observable<O>,
 ] {
-  const cache = new NestedMap<
-    A,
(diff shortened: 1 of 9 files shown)


packages/core/src/bind/connectFactoryObservable.ts
```diff
@@ -68,19 +68,23 @@ export default function connectFactoryObservable<A extends [], O>(
 
       return source$.subscribe(subscriber)
     }) as BehaviorObservable<O>
-    const reactObservable$ = reactEnhancer(publicShared$)
+    publicShared$.getValue = getEmpty
```
voliva, line 71: Why is it `getEmpty` by default instead of `sharedObservable$.getValue`? I think if no subscription happened yet, and someone calls `reactGetValue`, shouldn't it directly poke `sharedObservable$`?

packages/core/src/internal/useObservable.ts
```diff
@@ -1,56 +1,45 @@
...
-    ? current
-    : action
-
-const init = (source$: BehaviorObservable<any>) => source$.getValue()
+import { Observable } from "rxjs"
 
 export const useObservable = <O>(
-  source$: BehaviorObservable<O>,
+  source$: Observable<O>,
+  getValue: any,
```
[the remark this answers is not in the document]
josepot: Ups! Yep, thanks for spotting that :sweat_smile:

packages/core/src/bind/connectFactoryObservable.ts
josepot: sure... I mean, in that case `sharedObservable$.getValue will return `EMPTY_VALUE`, so it really is the same. But yeah, I'll make the change. Thanks!

voliva (approved):

PR #146, opened 2020-10-29
V0.6.0

josepot:
Changes of this release:

- Breaking change: `Subscribe` is now a super-set of `Suspense`. The default `fallback` is still `null`, though.

- Breaking change: We are no longer making the initial subscription on render. It's now the responsibility of the user to make sure that the initial subscription is there before the hook consumes the observable. That can be accomplished in many different ways, but the most common ones should be using `Subscribe` or with top-level subscriptions. In a nutshell what's going to happen is that if during render, if we have to throw a `Promise` to trigger suspense we won't create that promise unless the subscription of the Subject that's used for multicasting is already there. This guarantees that no side-effects will happen on render.

- Fix: If a "suspended" observable completes without emitting any values, we will throw an error that can be captured from an error-boundary.

- Fix: If a `Subscribe` component gets unmounted while one of its children is on Suspense, the underlying observable will be disposed.


packages/core/src/Subscribe.tsx
```diff
@@ -12,15 +17,20 @@ import { Observable } from "rxjs"
...
 }> = ({ source$, children, fallback }) => {
   const [mounted, setMounted] = useState(0)
-  useEffect(() => {
-    const subscription = source$.subscribe()
+  useLayoutEffect(() => {
+    const subscription = source$.subscribe(noop, (e) =>
+      setMounted(() => {
+        throw e
+      }),
+    )
```
voliva, line 28: Brings up the question: what should happen on complete?

voliva (approved):

packages/core/src/Subscribe.tsx
josepot: Nothing, really. Also, most of the times won't even happen, because the streams returned by `bind` don't complete.

josepot:
Ok, let's get this in, then 🤞

PR #147, opened 2020-11-09
fix(core/subscribe): stop rendering children when source$ switches

voliva:
I had an issue caused because I was receiving the old value from a factory `bind(id => ...)` when switching its key based on a router's param.

We realised this shouldn't happen when using `Subscribe` boundaries properly, but Subscribe doesn't currently handle the case of the source stream switching.

In here I add two tests: The first one tries to represent the underlying issue in Subscribe, but it turned out really contrived. So I also added a second test which better represents the issue I was having initially.

### DIFF
--- a/packages/core/src/Subscribe.tsx
+++ b/packages/core/src/Subscribe.tsx
@@ -22,9 +22,9 @@ export const Subscribe: React.FC<{
   const [mounted, setMounted] = useState(() => {
     try {
       ;(source$ as any).gV()
-      return 1
+      return source$
     } catch (e) {
-      return e.then ? 1 : 0
+      return e.then ? source$ : null
     }
   })
   useLayoutEffect(() => {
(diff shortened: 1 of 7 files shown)


packages/core/src/Subscribe.tsx
```diff
@@ -33,11 +33,15 @@ export const Subscribe: React.FC<{
         throw e
       }),
     )
-    setMounted(1)
+    setMounted(source$)
```
josepot, line 36: Wouldn't it be better to set the state in a conditional clause on render? :thinking: 

Something, like this?

```tsx
const init = (source$: Observable<any>): [boolean, Observable<any>] => {
  const result = [true, source$] as const
  try {
    ;(source$ as any).gV()
  } catch (e) {
    if (!e.then) result[0] = false
  }
  return result
}

export const Subscribe: React.FC<{
  source$: Observable<any>
  fallback?: NonNullable<ReactNode> | null
}> = ({ source$, children, fallback }) => {
  const [state, setState] = useState(() => init(source$))

  if (state[1] !== source$) setState(init(source$))

  useLayoutEffect(() => {
    const subscription = source$.subscribe(noop, (e) =>
      setState(() => {
        throw e
      }),
    )
    setState(current => current[0] ? current : [true, current[1]])
    return () => {
      subscription.unsubscribe()
    }
  }, [source$])

  return (
    <Suspense fallback={fallback || null}>
      {state[0] ? children : <Throw />}
    </Suspense>
  )
}
```

josepot:
Thanks a lot for this @voliva ! I've added a bunch of commits with tiny improvements. Please, could you please re-review it before we merge?

PR #149, opened 2020-12-22
fix(core/bind): resulting observable of bind should emit the default …

josepot:
…value if provided

I really think that we should consider this a bug-fix, although someone could argue that it's a breaking change... @voliva what do you think?


voliva (approved):
That makes sense - somehow it feels the expected behaviour should be the one that this PR adds (otherwise you need to add a redundant-looking `startWith`).

PR #155, opened 2021-01-07
fix(bind): supports observables of functions (0.5)

### LINKED ISSUE #152: When an Observable emits a function
Upgrading from `@react-rxjs/core@0.2.0` -> `@react-rxjs/core@0.5.0` came across this regression. 

When an observable returns a function, we get a stringified (plus evaluated!) version of the function. This worked in the prior version.

```ts

const counter$ = timer(0, 1000).pipe(startWith(0), shareReplay(1));

const [useBoundMethodThatDoesntWork] = bind(() =>
  counter$.pipe(
    map((count) => {
      return (s) => `${s}: ${count}`;
    })
  )
);

// .....


const method = useBoundMethod();
method('a') // -> Not a function
typeof method // -> 'stirng'
return ({method})
  // -> function (s) { return "".concat(s, ": ").concat(count); }: 0:
```
[Code](https://codesandbox.io/s/observable-that-returns-a-function-2vilq?file=/src/App.js)

voliva:
closes #152

### DIFF
--- a/packages/core/src/bind/connectFactoryObservable.test.tsx
+++ b/packages/core/src/bind/connectFactoryObservable.test.tsx
@@ -1,23 +1,29 @@
 import {
-  from,
-  of,
-  defer,
-  concat,
-  BehaviorSubject,
-  throwError,
-  Observable,
-  Subject,
-} from "rxjs"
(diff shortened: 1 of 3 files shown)


josepot (approved):
Awesome, thanks @voliva !

PR #158, opened 2021-01-10
fix(bind): supports variadic number of arguments

### LINKED ISSUE #156: Rest parameters on >0.4.3 and <0.6.1
This is fixed in `0.6.1` so just looking for advice on this, and whether a fix can be patched into `0.5.x`
Affects versions >0.4.3 <0.6.1

```ts
const [useRequiresTheArray] = bind((s: string, a: string[]) => {
  return of(a.join(s));
});
```

is fine, but

```ts
const [useRequiresTheArray] = bind((s: string, ...a: string[]) => {
  return of(a.join(s));
});
```
```text
Warning: The final argument passed to useEffect changed size between renders. The order and size of this array must remain constant.
```

[Example Code](https://codesandbox.io/s/upbeat-driscoll-d8knk?file=/src/App.js)

voliva:
Closes #156 

I'm not really sure we would officially support this feature, although it has been working on and off for a while. In our mental model, the parameters of a parametric `bind` are used to identify an instance of an entity.

As v0.5 is a version that helps the migration from 0.4 to 0.6 I think it's useful anyway to have this in.

### DIFF
--- a/packages/core/src/bind/connectFactoryObservable.test.tsx
+++ b/packages/core/src/bind/connectFactoryObservable.test.tsx
@@ -59,6 +59,22 @@ describe("connectFactoryObservable", () => {
       expect(result.current).toBe(3)
     })
 
+    it("accepts a changing number of arguments", async () => {
+      const [useConcatNumber] = bind((...args: number[]) => of(args.join(",")))
+      const { result, rerender } = renderHook(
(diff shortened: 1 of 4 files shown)


josepot:
Thanks a lot for this @voliva !

PR #159, opened 2021-01-11
V0.6.2

### LINKED ISSUE #157: Component receiving and empty object when remounting
Steps to reproduce:

1. Call bind with an observable that emits asynchronously (e.g. 1000ms)
2. Create a component that uses the hook from (1)
3. Use this component in two different routes, each with its own `<Subscribe>` boundary
4. Switch from one route to the other

Expected behaviour:

* The hook returns the actual value after suspending.

Current behaviour (0.6.1):

* The hook returns `{}` (EMPTY_VALUE) without suspending.

Sandbox: https://codesandbox.io/s/react-rxjs-empty-value-0n7zj?file=/src/App.tsx

josepot:
In this PR:
- remove warnings from the factory-bind test
- closes #157 
- upgrade dependencies
- @react-rxjs/core@0.6.2 patch

### DIFF
--- a/package.json
+++ b/package.json
@@ -40,11 +40,11 @@
     "trailingComma": "all"
   },
   "devDependencies": {
-    "cross-env": "^7.0.2",
-    "eslint-plugin-prettier": "^3.1.4",
-    "husky": "^4.3.0",
-    "lint-staged": "^10.5.1",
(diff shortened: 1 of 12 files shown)


voliva (approved):
🚀

PR #168, opened 2021-02-27
Utils improvements

josepot:
Lots of stuff in this PR:

- Get rid of the `CollectedObservable`
- Add the `geGroupedObservable` function, instead (I'm very open to suggestions for a better name :see_no_evil: )
- Get rid of the optional argument `filter` of the `collect` operator
- For a matter of performance and overall bundle-sizes: use a lot less rxjs operators


josepot:
I can't believe that the bundle size went down! :thinking:

packages/utils/src/collect.ts
```diff
@@ -1,82 +1,29 @@
...
+    | { t: CollectorAction.Set; k: K; v: GroupedObservable<K, V> }
+    | { t: CollectorAction.Delete; k: K }
+  >((observer) => {
+    const message = {
+      t: CollectorAction.Set as const,
+      k: source.key,
+      v: source,
+    }
+    observer.next(message)
+    ;(message as any).t = CollectorAction.Delete
```
voliva, line 15: smells like danger 😱

I see it currently doesn't hurt because the consumer of this enhancer reads the messages synchronously, but if it did asynchronously the initial message would get mutated from `Set` => `Delete`. You sure about this?

Edit - Thinking it better, yep it's not that important b/c it's an internal implementation - If it ever breaks the external behaviour, tests would catch that

packages/utils/src/internal-utils.ts
```diff
@@ -50,22 +37,44 @@ export const collector = <K, V, VV>(
...
+                } else {
+                  map.delete(v.k)
+                }
+                emitted = true
+                observer.next(map)
+              },
+              (e) => {
+                observer.error(e)
+              },
+            ),
```
voliva, line 65: I'm thinking that both collect and collectValues delete the key only on complete or error (except that in collectValues it propagates the error too) - Maybe this enhancer thing could be further simplified if it only emits `CollectorAction.Set` events instead?

e.g.
```ts
enhancer(x).subscribe(
   ([k,v]) => {
      map.set(k,v);
      emitted = true
      observer.next(map)
   },
   e => {
     observer.error(e)
   },
   () => {
      map.delete( /*.... errr where do I get the key from? 😅 */)
   }
``` 

Mhh maybe not 😅 - Or yes, isn't the `x ` a GroupObservable? That would have the key.

josepot: That's a very good point... Now that we got rid of the `filter` of `collect`, this can be simplified quite a lot. Let me think about it, thanks for pointing this out!

josepot: Check the latest commit! Thanks a lot for this comment!

packages/utils/src/collect.ts
josepot: I realize that we no longer need the `CollectorAction` enum :partying_face:

voliva (approved):

PR #174, opened 2021-03-30
feat(utils): partitionByKey && mergeActiveKeys

josepot:
Hopefully this will allow us to deprecate the following utils:
- `split`
- `collect`
- `collectValues`

While also providing a simpler way to create subscriptions for parametric bind Observables.

@voliva could you please finish this?


josepot:
Code coverage went down to 99.53%... alt text

josepot:
I know that we've had this conversation "offline" before, but I really think that the order of the return value of `partitionByKey` should be reversed: first the inner-observable-getter and then the observable of keys... :thinking: 
image

voliva (approved):
👍 

Combinekeys test has a typo: emmit vs emit

josepot:
> Combinekeys test has a typo: emmit vs emit

Fixed! Thanks!

PR #175, opened 2021-03-30
chore: upgrade dependencies

josepot:


josepot:
The total bundle-size has increased +2.5% just by upgrading the dependencies!? We need to look into this :thinking: cc @voliva

PR #181, opened 2021-04-01
feat(utils): `createKeyedSignal` improve mapper overload

josepot:
I think that `createKeyedSignal` is by far one of the best additions that we've made to the `utils` package. However, the "mapper" overload is not very ergonomic, for 2 reasons:

- When writing the key selector, TS hasn't been able to infer the type returned by the mapper yet
- In many cases what's desirable is to create an Observable of a mapped value that doesn't have the key

That's why I think that overload should be improved. Unfortunately, that implies a breaking change :confused:, luckily, though, we are still under major version zero :see_no_evil:

### DIFF
--- a/packages/utils/src/contextBinder.ts
+++ b/packages/utils/src/contextBinder.ts
@@ -1,11 +1,6 @@
 import { Observable } from "rxjs"
 import { bind } from "@react-rxjs/core"
-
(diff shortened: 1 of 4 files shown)


josepot:
Closing this in favor of #182 , since #182 accomplishes the same thing without breaking changes.

PR #183, opened 2021-04-01
feat(core): factory `bind` accept function for default value

josepot:
I've been meaning to make this improvement for a while now, but I never got to it... Technically speaking, this is a breaking change. Because in the super-rare event that someone was using a factory `bind` with Observables of functions and that the default-value overload was being used, then they will have to wrap that default function into another function... So, yeah, this should go into `0.7.0` I guess :shrug: 

@voliva  what do you think?


voliva (approved):
go for it 🚀

PR #191, opened 2021-04-16
feat(core): 🎉 Remove source$ prop from Subscribe 🎉

josepot:
I won't be able to sleep tonight because of this... With this change I finally feel like we are ready for v1.0.0! :tada:


josepot:
Closing this in favor of #192

PR #192, opened 2021-04-16
feat(core): 🎉 Subscribe: Make source$ optional 🎉

josepot:


packages/core/src/Subscribe.tsx
```diff
@@ -1,58 +1,103 @@
...
-  }
-}
-
 /**
- * A React Component that creates a subscription to the provided observable once
- * the component mounts and it unsubscribes when the component unmounts.
+ * A React Component that:
+ * - collects the subscriptions of its children and it unsubscribes them when
+ * the component unmounts.
+ * - if a source$ property isused, then it ensures that the subscription to the
```
voliva, line 25:
```suggestion
 * - if a source$ property is used, then it ensures that the subscription to the
```

packages/core/src/internal/share-latest.ts
```diff
@@ -91,7 +91,18 @@ const shareLatest = <T>(
...
+      let err = EMPTY_VALUE
+      const sub = result.subscribe({
+        error(e) {
+          err = e
+        },
+      })
+      if (err !== EMPTY_VALUE) throw err
+
+      outterSubscription.add(sub)
+      return result.gV()
```
voliva, line 105: I suspect if `source$` emits complete synchronously this could result in an infinite loop... I'll need to check

voliva (commented):
Awesome! Let's try it out and see how it plays... this completely changes the game 🎉

packages/core/src/internal/share-latest.ts
voliva: Just checked, all good: this subscription doesn't emit the inner complete, so it doesn't unsubscribe, keeping the refCount > 0 and preventing the currentValue from being reset.

josepot: No, but this is a good point, b/c before we go ahead and publish v1.0.0 I really want to revisit this decision of preventing the "enhanced" observable from completing...

packages/core/src/Subscribe.tsx
josepot: Thanks!

PR #194, opened 2021-05-06
fix(shareLatest): properly closing sync observables

josepot:
Basically, bringing [this](https://github.com/ReactiveX/rxjs/pull/6370) to react-rxjs


packages/core/src/internal/share-latest.ts
```diff
@@ -49,29 +65,14 @@ const shareLatest = <T>(
           subject!.complete()
         },
       )
-      if (subscription.closed) subscription = null
```
josepot, line 52: No longer needed! :tada:

voliva: good one

voliva (approved):

PR #199, opened 2021-05-23
feat(utils/combineKeys): emit set with changed keys

voliva:
Recently I found a use case where I'd like to have an operator similar to `combineKeys`, but that instead of emitting a map of values it would just emit the latest individual change.

I couldn't find any way of composing the existing operators to do this, so I had to make a custom operator which had mostly combineKeys' code changing a few lines.

After discussing it we found out that this operator would be easy to implement if the existing `combineKeys` also emitted the list of keys that have changed. This is a posible implementation.

With this operator, the mergeKeys operator I had in mind would become something like:

```ts
const mergeKeys = <K, T>(keys$: Observable<K>, getInner$: (key: K) => Observable<T>): Observable<T> =>
  combineKeys(keys$, getInner$).pipe(
    mergeMap(values => {
      return Array.from(values.changes)
        .map(key => values.get(key))
        .filter(value => value !== undefined)
    })
  )
```

Doubts on this one:
- Is it good to just add a property into the `Map`?
- I went for the same pattern to make a copy of the Set... I think an alternative would be to just return the existing Set and instantiate a new one for the next emission

### DIFF
--- a/packages/utils/src/combineKeys.test.ts
+++ b/packages/utils/src/combineKeys.test.ts
@@ -204,4 +204,60 @@ describe("combineKeys", () => {
       })
     })
   })
+
+  describe("change set", () => {
+    it("contains all of the keys initially present in the stream", () => {
+      scheduler().run(({ expectObservable }) => {
+        const keys = concat(of(["a", "b", "c"]), NEVER)
+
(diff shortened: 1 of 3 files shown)


josepot (approved):
Awesome! Thanks!

peicodes:
Can't wait to use this!

PR #202, opened 2021-05-29
fix(partitionByKey): unsubscribe from all streams when refcount reaches 0

voliva:
I've found a couple of issues with partitionByKey:

- When the source completes, it completes the key$ and every instance observable. I think it should delegate this to the projection function so that when it completes, then the key is removed.
- After everyone unsubscribed from any of the streams it returns, it kept a subscription open to the inner observables.

I couldn't manage to change split/collect to solve these issues without breaking other split/collect tests, so I've reimplemented it as a standalone operator. I added a bunch of tests for added coverage, but the relevant ones are in these commit: https://github.com/re-rxjs/react-rxjs/commit/717b5dc4df853808e696147fa6cbfb7632971c20

- it keeps the existing keys alive when the source completes
- it unsubscribes from all streams when refcount reaches 0
- it emits a complete on the inner observable when the source completes
- it emits an error when the source errors

Edit: with this change we've decided to remove the old split/collect/collectValue methods


josepot (approved):
I love it! Thanks a lot for this @voliva !

PR #223, opened 2021-09-21
patch: upgrade dev dependencies & patch

josepot:


voliva (approved):
👍 

The increase of the bundle size visible on bundlewatch is due to the new version of es-build, where it adds an adittional check for cjs builds: https://github.com/evanw/esbuild/pull/1291

It only effects the cjs sizes, adding ~100 bytes to each package, code looks like:

```js
var __require = typeof require !== "undefined" ? require : (x) => {
  throw new Error('Dynamic require of "' + x + '" is not supported');
};
```

I think it's fine

PR #234, opened 2022-02-05
refactor: use syncExternalStore in preparation for react@18

voliva:
This is a quick reimplementation, but I'm not fully convinced it's implemented the way it should be implemented. Don't merge, needs more thought

### DIFF
--- a/packages/core/src/internal/share-latest.ts
+++ b/packages/core/src/internal/share-latest.ts
@@ -144,6 +144,13 @@ const shareLatest = <T>(
       subscription!.add(pSubs)
     }))
   }
+  result.sB = (callback) => {
+    const sub = result.subscribe({
+      next: callback,
+      error: callback,
+    })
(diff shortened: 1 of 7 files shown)


packages/core/src/internal/share-latest.ts
```diff
@@ -132,6 +144,24 @@ const shareLatest = <T>(
...
+  }
+
+  const cachedCallback: WeakMap<Subscription, () => any> = new WeakMap()
+  result.gVS = (subscription) => {
+    if (!subscription) {
+      return result.gV
+    }
+    if (!cachedCallback.has(subscription)) {
+      cachedCallback.set(subscription, () => result.gV(subscription))
+    }
```
voliva, line 162: I'm now thinking this is creating a hard reference to `subscription`, so it won't get GCed. Will need more rework.

voliva:
Closed in favour of #236

PR #236, opened 2022-03-10
Use sync external store (and moar)

josepot:
A few things in this PR:
- leverage `useSyncExternalStore`
- remove the hacky `setTimeout` for when a suspended hook throws
- upgrade our devDependencies
- add `"type": "module"` to all the `package.json`
- accordingly patch the 3 libraries


voliva (approved):
🚀

PR #238, opened 2022-03-20
v0.9.0 rxjs-state integration

josepot:
We finally got a proper integration with `rxjs-state` which works with both React 17 and React 18, and that it's also backwards compatible with v0.8.x :tada: I really think that this is what version 1 is going to look like. Although, let's first publish v0.9.0, b/c I'm not 100% sold on whether we should keep `SUSPENSE` in its current form.

I've put @voliva as a co-author of this commit b/c this really has been a joined effort.

What's this change is about:

- `@react-rxjs/core` becomes a superset of `@josepot/rxjs-state` (which hopefully in the near future will be under `@rxjs-state/core` :crossed_fingers:)
- Added new `useStateObservable` hook, which can be used for consuming the `StateObservables that come from `rxjs-state` :tada: 
- We are keeping `shareLatest` for backwards compatible reasons. However, now it's just a wrapper on RxJS `share`, which is why the rxjs peer-dependency range has changed.
- the tests now pass on React 18
- a bunch of improvements and optimizations on the internals


voliva (approved):

packages/core/src/bind/index.ts
```diff
@@ -23,7 +24,7 @@ import { EMPTY_VALUE } from "../internal/empty-value"
 export function bind<T>(
   observable: Observable<T>,
   defaultValue?: T,
-): [() => Exclude<T, typeof SUSPENSE>, Observable<T>]
+): [() => Exclude<T, typeof SUSPENSE>, StateObservable<T>]
```
voliva, line 27: Should we expose the `DeafultedStateObservable` with the defaultValue overload on this release?

voliva: I pushed a commit adding them in

packages/dom/src/batchUpdates.test.tsx
```diff
@@ -99,7 +99,7 @@ describe("batchUpdates", () => {
     }
   })
 
-  test("it triggers nested updates when batchUpdates is not used", async () => {
+  test.skip("it triggers nested updates when batchUpdates is not used", async () => {
```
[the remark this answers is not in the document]
josepot: yep... it turns out that React 18 finally batches updates OOTB, so I don't quite understand why it keeps exposing `unstable_batchedUpdates` :shrug:

packages/core/src/bind/index.ts
josepot: Awesome!

PR #252, opened 2022-03-31
feat: utils `toKeySet`

voliva:


packages/utils/src/toKeySet.test.ts
```diff
@@ -0,0 +1,99 @@
+import { KeyChanges } from "./partitionByKey"
+import { TestScheduler } from "rxjs/testing"
+import { toKeySet } from "./toKeySet"
```
josepot, line 3:
```suggestion
import { toKeySet } from "./"
```

packages/utils/src/toKeySet.test.ts
```diff
@@ -0,0 +1,99 @@
...
+const scheduler = () =>
+  new TestScheduler((actual, expected) => {
+    expect(actual).toEqual(expected)
+  })
+
+describe("toKeySet", () => {
+  it("transforms key changes to a Set", () => {
+    scheduler().run(({ expectObservable, cold }) => {
+      const expectedStr = "                     xe--f-g--h"
+      const source$ = cold<KeyChanges<string>>("-a--b-c--d", {
```
josepot, line 15: Silly, but let's make sure that errors get propagated (you will see why in the next comment :see_no_evil:)
```suggestion
      const expectedStr = "                     xe--f-g--h#"
      const source$ = cold<KeyChanges<string>>("-a--b-c--d#", {
```

packages/utils/src/toKeySet.ts
```diff
@@ -0,0 +1,26 @@
...
+            for (let k of changes.keys) {
+              acc.delete(k)
+            }
+          }
+
+          return acc
+        }, result),
+        defaultStart(result),
+      )
+    })
```
josepot, line 25: For something this trivial, I think it's best to create a new Observable rather than using 3 different operators...

```suggestion
     new Observable<Set<K>>((observer) => {               
       const result = new Set<K>()                        
       let pristine = true                                
       const subscription = source$.subscribe({           
         next({ type, keys }) {                           
           const action = type === "add" ? type : "delete"
           for (let k of keys) {                          
             result[action](k)                            
           }                                              
           observer.next(result)                          
           pristine = false                               
         },                                               
         error(e) {                                       
           observer.error(e)                              
         },                                               
         complete() {                                     
           observer.complete()                            
         },                                               
       })                                                 
       if (pristine) observer.next(result)                
       return subscription                                
     })                                        
```

josepot (approved):
Thanks @voliva! Please consider my comments before merging :pray:

PR #279, opened 2022-08-30
fix Subscribe throwing NoSubscribersError on synchronous completes

voliva:
Unfortunately the test doesn't reproduce it :(

Maybe it's related to the testing framework? A simple sandbox does reproduce it: https://codesandbox.io/s/react-rxjs-playground-sfty1h?file=/src/App.tsx

I'll take a closer look this evening

### DIFF
--- a/packages/core/src/Subscribe.test.tsx
+++ b/packages/core/src/Subscribe.test.tsx
@@ -1,4 +1,9 @@
-import { sinkSuspense, state, SUSPENSE } from "@rx-state/core"
+import {
+  EmptyObservableError,
+  sinkSuspense,
+  state,
+  SUSPENSE,
+} from "@rx-state/core"
 import { act, render, screen } from "@testing-library/react"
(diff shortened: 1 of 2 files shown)


packages/core/src/Subscribe.test.tsx
```diff
@@ -355,6 +361,46 @@ describe("Subscribe", () => {
...
+          <Subscribe fallback={<div>Loading...</div>}>
+            <ErrorComponent />
+          </Subscribe>
+        </TestErrorBoundary>,
+      )
+
+      // Can't have NoSubscribersError
+      // Can't have "Cannot update component (`%s`) while rendering a different component"
+      globalErrors.mock.calls.forEach(([errorMessage]) => {
+        console.log(errorMessage)
```
voliva, line 387: console.log

voliva, line 387:
```suggestion
```

josepot (approved):

PR #303, opened 2023-07-11
Upgrade dependencies, migrate to vitest

voliva:
We were on a quite old version of typescript, jest, etc. so I've tried to upgrade them.

However, I've found a lot of issues trying to get jest to work. Apparently their support for ESModules is still "experimental" and it's a nightmare to set up (specially with typescript in-between).

After trying for an hour I've decided to migrate it to vitest, which is way more out-of-the-box, specially with typescript support.

The only "drawback" is that we need to import the describe/it functions. In theory we don't actually need it because I've set up `globals: true`, because react-testing-library needs that, but I don't know how to tell typescript that these globals are available only on test files. So anyway, I think it doesn't hurt that much.

Accidentally closes #290 and #294


josepot (approved):
Thanks a lot!

PR #306, opened 2023-08-04
Add getServerSnapshot, fix loop on SSR Subscribe

### LINKED ISSUE #305: NextJS 13 getServerSnapshot issue
When using react-rxjs inside a nextjs client component, I see this on the server sometimes:

```
- error Error: Missing getServerSnapshot, which is required for server-rendered content. Will revert to client rendering.
```

There's a similar issue on zustand: https://github.com/pmndrs/zustand/issues/886.

I only saw this error when I started using `react-rxjs`, seems to apply to state/hook related libraries? I'm not entirely sure why it would occur when I start using `react-rxjs`.

voliva:
Fixes #305 

Something important to keep in mind is that in SSR any children within a `Subscribe` will not render, because given that components never mount in SSR (including `<Subscribe>`), it will never get into the stage of rendering the children.

And `StateObservable` with top-level subscriptions will have their state shared between clients.

### DIFF
--- a/packages/core/src/useStateObservable.ts
+++ b/packages/core/src/useStateObservable.ts
@@ -14,7 +14,11 @@ type VoidCb = () => void
 
 interface Ref<T> {
   source$: StateObservable<T>
-  args: [(cb: VoidCb) => VoidCb, () => Exclude<T, SUSPENSE>]
+  args: [
+    (cb: VoidCb) => VoidCb,
+    () => Exclude<T, typeof SUSPENSE>,
+    () => Exclude<T, typeof SUSPENSE>,
+  ]
 }
 
 export const useStateObservable = <O>(
@@ -46,7 +50,7 @@ export const useStateObservable = <O>(
 
     callbackRef.current = {
       source$: null as any,
-      args: [, gv] as any,
+      args: [, gv, gv] as any,
     }
   }
 
--- a/package-lock.json
+++ b/package-lock.json
@@ -12,6 +12,7 @@
         "@babel/preset-env": "^7.22.7",
         "@babel/preset-typescript": "^7.22.5",
         "@testing-library/react": "^14.0.0",
+        "@types/node": "^20.4.7",
         "@types/react": "^18.2.14",
         "@types/react-dom": "^18.2.6",
         "@vitest/coverage-v8": "^0.33.0",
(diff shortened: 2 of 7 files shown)


packages/core/src/useStateObservable.ts
```diff
@@ -46,7 +50,7 @@ export const useStateObservable = <O>(
 
     callbackRef.current = {
       source$: null as any,
-      args: [, gv] as any,
```
voliva, line 49: I think we can use `gv` for serverSnapshot because everything still matches:

```ts
    const gv: <T>() => Exclude<T, typeof SUSPENSE> = () => {
      const src = callbackRef.current!.source$ as DefaultedStateObservable<O>
      if (!src.getRefCount() && !src.getDefaultValue) {
        // `subscription` will always be null, because it's impossible for a component to run this hook if it's inside a <Subscribe>
        // Then this will happen only if a component without a <Subscribe> is using a StateObservable that doesn't have a subscription => Missing Subscribe
        if (!subscription) throw new Error("Missing Subscribe!")
        subscription(src)
      }
      // The rest should also work:
      // Observables with top-level subscriptions will return their value, or trigger Suspense
      // Observables without subscriptions and defaultValue will return their defaultValue
      return getValue(src)
    }
```

packages/core/src/useStateObservable.ts
```diff
@@ -46,7 +50,7 @@ export const useStateObservable = <O>(
...
+      args: [, gv, gv] as any,
```
voliva, line 53: I think we can use `gv` for serverSnapshot because everything still matches:

```ts
    const gv: <T>() => Exclude<T, typeof SUSPENSE> = () => {
      const src = callbackRef.current!.source$ as DefaultedStateObservable<O>
      if (!src.getRefCount() && !src.getDefaultValue) {
        // `subscription` will always be null, because it's impossible for a component to run this hook if it's inside a <Subscribe>
        // Then this will happen only if a component without a <Subscribe> is using a StateObservable that doesn't have a subscription => Missing Subscribe
        if (!subscription) throw new Error("Missing Subscribe!")
        subscription(src)
      }
      // The rest should also work:
      // Observables with top-level subscriptions will return their value, or trigger Suspense
      // Observables without subscriptions and defaultValue will return their defaultValue
      return getValue(src)
    }
```

josepot (approved):
Nice! Thanks a lot @voliva !

CMCDragonkai:
Does this means one should not use this in server side components?

voliva:
> Does this means one should not use this in server side components?

At the moment I don't really have an answer for this, I haven't played that much with SSR or server-side components.

It's just important to keep in mind that in server-side:

1. Anything inside `<Subscribe>` will not render. `<Subscribe>` is used to subscribe to a source$ before the children are rendered. If `<Subscribe>` can't subscribe to the source because it never gets mounted, then it will never render its children.
2. `StateObservable`s with defaultValue will return their default value (just as normal).
3. Any `StateObservable`s with a subscription somewhere else (e.g. a top-level subscription) will have their state shared between every render. This could be useful, or it could be annoying to get around.

My best advice (again, without having played that much with SSR/NextJS) is to have every StateObservable with a defaultValue on those components that will run on the Server. It gives the most predictable and simple results: They will return the defaultValue while on the server, and execute the subscription when on the client.

Maybe on a future we can leverage something like https://github.com/reactjs/rfcs/pull/229 so that server components also subscribe to the observables, but there are some serious questions regarding cancellation and state sharing.

PR #315, opened 2024-01-17
Add comments and questions

bruceharris:
Added comments, a question, and made a few minor changes to aid my own understanding of the code. If there's anything incorrect about the comments I've added, feedback/clarification would be greatly appreciated. Happy to iterate on this to evolve this PR to a point where it can be merged so the comments can be helpful to other readers, but at present I'm primarily concerned with verifying my own understanding. Thanks in advance!


josepot:
Please use the [Discussion](https://github.com/re-rxjs/react-rxjs/discussions) section of the project for asking questions.

PR #321, opened 2025-01-09
Update to React 19

rveciana:
Upgrade the dependencies for the project
Fix broken tests because of the vitest upgrade
Fix small errors because of the react upgrade


packages/core/src/bind/connectFactoryObservable.test.tsx
```diff
@@ -84,13 +84,17 @@ describe("connectFactoryObservable", () => {
 
       render(<TestSuspense />)
 
-      expect(screen.queryByText("Result")).toBeNull()
-      expect(screen.queryByText("Waiting")).not.toBeNull()
+      vi.waitFor(() => {
+        expect(screen.queryByText("Result")).toBeNull()
+        expect(screen.queryByText("Waiting")).not.toBeNull()
+      })
```
voliva, line 90: From what I see, there are only 3 tests that are failing some of these:
- connectObservable suspends the component when the observable starts emitting suspense.
- connectObservable correctly unsubscribes when the Subscribe component gets unmounted.
- connectFactoryObservable hook suspends the component when the observable hasn't emitted yet.

Instead of wraping every assertion with a `waitFor`, I'd rather have them controlled. This is a change on behaviour that we should probably fix, or at least have identified. It doesn't seem like a big behaviour change if it's something now asynchronous, but it's better to dig down on that instead of wraping everything with `waitFor` indiscriminately.

voliva (commented):
Thanks @rveciana for the contribution!

I see you have updated the `devDependencies` of the monorepo, but the `@react-rxjs/core` pacakge is left untouched, aside from linting/prettier changes.

I think we have to update the peerDependency of `@react-rxjs/core` to declare it's compatible with react@19

packages/core/src/bind/connectFactoryObservable.test.tsx
rveciana: I removed all the unnecessary waitFor and added a timeout to the remaining one so the tests are not flaky as in the first version.
The remainig waitFor seem necessary not to increase the await wait(110) value, that should have a meaningful time for the logic, not a time to make the test pass.

voliva (commented):
Thanks again @rveciana!

It looks good to me, but you have removed some generics on the test files that gives me TS errors on my IDE - I tried adding a commit fixing those to this PR but I couldn't, you can cherry-pick it if you want: https://github.com/re-rxjs/react-rxjs/commit/69233017c9aeba3be181ef978b7f1494cd05592a

The CI is failing for an unrelated issue with some code coverage tool we have, I'll take a look as a separate issue.

rveciana:
Thanks for finsding this. A wrong vscode configuration did it.

voliva (approved):

voliva:
@all-contributors please add @rveciana for maintenance

voliva:
Published as @react-rxjs/core@0.10.8

